PR #13, opened 2016-07-08
Non-blocking binary writer.

tgregg:


amazon/ion/core.py
```diff
@@ -154,6 +154,25 @@ def derive_annotations(self, annotations):
...
+        """Derives a new event from this one setting the ``value`` attribute.
+
+        Args:
+            value: (any):
+                The value associated with the derived event.
+
+        Returns:
+            IonEvent: The newly generated event.
+        """
+        return IonEvent(
```
almann: Given `IonThunkEvent` we probably should be explicit that this creates a non-thunk event.

amazon/ion/writer_binary.py
```diff
@@ -0,0 +1,221 @@
...
+# Python 2/3 compatibility
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from .exceptions import IonException
+from .symbols import SID_ION_SYMBOL_TABLE, SID_IMPORTS, SHARED_TABLE_TYPE, \
+    SID_NAME, SID_VERSION, SID_MAX_ID, SID_SYMBOLS, SymbolTable, LOCAL_TABLE_TYPE
+from .writer_binary_raw import _raw_binary_writer, _WRITER_EVENT_NEEDS_INPUT_EMPTY,\
+    _WRITER_EVENT_COMPLETE_EMPTY
```
almann: Let's organize the imports a little (put all the constants together, and so on).

amazon/ion/writer_binary.py
```diff
@@ -0,0 +1,221 @@
...
+from .writer_buffer import BufferTree
+from .core import IonEventType, IonType, IonEvent, DataEvent, Transition
+from .util import coroutine, Enum, record
+from .writer import writer_trampoline, _drain, partial_transition, WriteEventType
+
+
+_IVM = bytearray([0xE0, 0x01, 0x00, 0xEA])
```
almann: Does this need to be `bytearray`, why not an immutable `bytes` sequence as `b'\xE0\x01\x00\xEA'`

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+
+import six
+
+from .writer_binary_raw_fields import _write_varuint, _write_uint, _write_varint, _write_int
+from .core import IonEventType, IonType, DataEvent, Transition
+from .util import coroutine, Enum
+from .writer import writer_trampoline, partial_transition, WriteEventType
+
+
+class _TypeIds(Enum):
```
almann: We should probably consolidate this with the ones in the `binary_reader` at some point.  Added #14 to track this.

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+    decimal = 0x50
+    timestamp = 0x60
+    symbol = 0x70
+    string = 0x80
+    clob = 0x90
+    blob = 0xA0
+    list = 0xB0
+    sexp = 0xC0
+    struct = 0xD0
+    annotation_wrapper = 0xE0
```
almann: I would prefer it if our enumeration constants were all caps.

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+    Notes:
+        Blob, clob, list, and sexp also have single-octet encodings, but the current implementation handles these
+        implicitly.
+    """
+    int = _TypeIds.pos_int | 0x0  # Int zero is always encoded using the positive type ID.
+    float = _TypeIds.float | 0x0  # Represents zero.
+    decimal = _TypeIds.decimal | 0x0  # Represents 0d0.
+    string = _TypeIds.string | 0x0  # Represents the zero-length (empty) string.
+    symbol = _TypeIds.symbol | 0x0  # Represents symbol zero.
+    struct = _TypeIds.struct | 0x0  # Represents a struct with zero fields.
```
almann: Isn't the above redundant?  The types are all defined in terms of an octet with the high-nibble set.

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+    _TypeIds.decimal | _NULL_INDICATOR,
+    _TypeIds.timestamp | _NULL_INDICATOR,
+    _TypeIds.symbol | _NULL_INDICATOR,
+    _TypeIds.string | _NULL_INDICATOR,
+    _TypeIds.clob | _NULL_INDICATOR,
+    _TypeIds.blob | _NULL_INDICATOR,
+    _TypeIds.list | _NULL_INDICATOR,
+    _TypeIds.sexp | _NULL_INDICATOR,
+    _TypeIds.struct | _NULL_INDICATOR
+]
```
almann: Probably another example for #14.

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+    else:
+        tid = _TypeIds.pos_int
+    _write_int_value(buf, tid, value)
+
+
+def _serialize_float(buf, ion_event):
+    float_value = ion_event.value
+    _validate_scalar_value(float_value, float)
+    # TODO Assess whether abbreviated encoding of zero is beneficial; it's allowed by spec.
+    if float_value.is_integer() and float_value == 0:
```
almann: technically this should be `float_value` == 0.0 to avoid coercion.

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+        buf.append(_Zeros.float)
+        return
+    # TODO Add an option for 32-bit representation (length=4) per the spec.
+    buf.append(_TypeIds.float | _LENGTH_FLOAT_64)
+    encoded = struct.pack('>d', float_value)  # '>' specifies big-endian encoding.
```
almann: No need to add a comment for `>`.

amazon/ion/writer_binary_raw.py
tgregg: Yes it is. I did this to be explicit because you had previously made a comment that it was confusing (though correct) to just write the type ID for the single-byte representation.

Is there some other way you'd prefer I address that confusion?

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+    buf.extend(encoded)
+
+
+def _write_decimal_value(buf, exponent, coefficient, sign=0):
+    length = _write_varint(buf, exponent)
+    if coefficient:  # The coefficient is non-zero, so the coefficient field is required.
```
almann: In general, these kinds of comments should really go on their own lines, in my not so humble opinion.

tgregg: Are you of the opinion that all inline comments should be on their own line?

amazon/ion/writer_binary_raw.py
```diff
@@ -0,0 +1,365 @@
...
+        length += _write_int(buf, coefficient)
+    elif sign:  # The coefficient is negative zero.
+        buf.append(_INT_NEG_ZERO)
+        length += 1
+    # Else the coefficient is positive zero and the field is omitted.
+    return length
+
+
+def digits_to_coefficient(digits):
```
almann: I was thinking here, does it make sense to scale the decimal and convert it to an integer instead of doing the base-10 conversion yourself?

E.g.

```python
d = Decimal(...)
sign, digits, exp = d.as_tuple()
coefficient = d.scaleb(-exp).to_integral_value()
```

amazon/ion/writer_binary_raw_fields.py
```diff
@@ -0,0 +1,240 @@
...
+# "Signed" - applies only to Int and VarInt - the signed fields (sign bit in first octet)
+# "Unsigned" - applies only to UInt and VarUInt - the unsigned fields (no sign bit)
+# "Variable" - applies only to VarInt and VarUInt - the variable-length fields (end bit required)
+# "Fixed" - applies only to Int and UInt - the fixed-length fields (no end bit)
+# "Int" - applies only to the Int field - the signed, fixed-length field
+# "VarInt" - applies only to the VarInt field - the signed, variable-length field
+# "UInt" - applies only to the UInt field - the unsigned, fixed-length field
+# "VarUInt" - applies only to the VarUInt field - the unsigned, variable-length field
+
+__VARIABLE_END_BIT_MASK = 0b10000000  # Both VarInt and VarUInt
```
almann: Some `__` to remove.

amazon/ion/writer_binary_raw_fields.py
```diff
@@ -0,0 +1,240 @@
...
+    - Decimal values with exponents and/or coefficients within the signed cache range will retrieve those fields from
+      the cache.
+    - All timestamps will retrieve their month, day, hour, minute, and second fields from the cache (otherwise it would
+      be an invalid datetime). The fractional fields may be cached if they meet the decimal value requirement from the
+      previous bullet.
+    - The first 64 symbols will have their symbol ID fields retrieved from the cache. This applies to symbol values,
+      field names, and annotations.
+    - If a value has less than 64 bytes of annotations (hopefully this is true for all values), then its annotation
+      wrapper's 'annotation length' field will be cached.
+    """
```
almann: Out of curiosity, did you measure a difference for doing this cache?

tests/writer_util.py
```diff
@@ -0,0 +1,92 @@
...
+_DT = datetime
+
+_E = IonEvent
+_IT = IonType
+_ET = IonEventType
+
+_STREAM_END_EVENT = (_E(_ET.STREAM_END),)
+
+
+class _P(record('desc', 'events', 'expected')):
```
almann: Now that this has a proper utility module--let's give it a proper name.

tests/writer_util.py
```diff
@@ -0,0 +1,92 @@
...
+from tests import is_exception
+from tests import noop_manager
+
+
+_D = Decimal
+_DT = datetime
+
+_E = IonEvent
+_IT = IonType
+_ET = IonEventType
```
almann: Any names you expect to have a user use should not have a `_` prefix, I am not sure which names fit the criteria.

tests/writer_util.py
```diff
@@ -0,0 +1,92 @@
...
+    if force_stream_end:
+        events += _STREAM_END_EVENT
+    return _P(
+        desc='SCALAR %s - %s' % (ion_type.name, expected),
+        events=events,
+        expected=expected,
+    )
+
+
+def _generate_scalars(scalars_map, force_stream_end=False):
```
almann: Looks like this is used outside of this module, let's delete the `_`.

tests/test_writer_text.py
```diff
@@ -17,43 +17,24 @@
...
-def _scalar_p(ion_type, value, expected):
-    return _P(
-        desc='SCALAR %s - %s' % (ion_type.name, expected),
-        events=(_E(_ET.SCALAR, ion_type, value),),
-        expected=expected,
-    )
+from amazon.ion.symbols import SymbolToken
+from amazon.ion.writer import blocking_writer
+from amazon.ion.writer_text import raw_writer
+from tests.writer_util import assert_writer_events, _E, _ET, _DT, _IT, _D, _P, _generate_scalars, _generate_containers
```
almann: I wouldn't shorthand these in the utility module, I'd give them proper names and short hand them in the tests.

amazon/ion/writer_binary_raw.py
almann: Can you just elide the `| 0x0`?

amazon/ion/writer_binary_raw.py
almann: No, but if the comment refers to a statement or block, I generally think it goes on its own line.  I generally use inline comments to annotate something about a part of a line.

amazon/ion/core.py
tgregg: Added to comment.

amazon/ion/writer_binary.py
tgregg: Not sure exactly what you want here, but I grouped the constants together and alphabetized the locations.

amazon/ion/writer_binary.py
tgregg: Done.

amazon/ion/writer_binary_raw.py
tgregg: Done.

amazon/ion/writer_binary_raw.py
tgregg: Done.

amazon/ion/writer_binary_raw.py
tgregg: Done.

amazon/ion/writer_binary_raw.py
tgregg: Removed.

amazon/ion/writer_binary_raw.py
tgregg: Done.

amazon/ion/writer_binary_raw.py
tgregg: Good call. That simplifies the code and performs a little better.

amazon/ion/writer_binary_raw_fields.py
tgregg: Done.

amazon/ion/writer_binary_raw_fields.py
tgregg: Yes. After looking at some profiling results generated by my (admittedly limited) test data, I found that field writing was taking the most time. The cache reduced the amount of time spent there by around 50%, leading to a modest 10% latency reduction. It's not huge, but I at least want to keep it around until we focus more on benchmarking, when we can revisit its merits if we wish. Using the rationale I put together in the comment, I convinced myself it would probably be widely beneficial, but I'm open to being proven wrong and reducing the size of the code base.

tests/writer_util.py
tgregg: In line with your other comment, I moved these back into the writer tests where they are used.

tests/writer_util.py
tgregg: Done.

tests/test_writer_text.py
tgregg: Removed from the utility module and moved back to the tests where they are used.

amazon/ion/core.py
```diff
@@ -122,7 +122,7 @@ def derive_field_name(self, field_name):
         Args:
             field_name (Union[amazon.ion.symbols.SymbolToken, unicode]): The field name to set.
         Returns:
-            IonEvent: The newly generated event.
+            IonEvent: The newly generated non-thunk event.
```
almann: Actually if you look below the generated event is based on the event derived, so it could be a thunk event.

amazon/ion/core.py
```diff
@@ -142,7 +142,7 @@ def derive_annotations(self, annotations):
                 The annotations associated with the derived event.
 
         Returns:
-            IonEvent: The newly generated event.
+            IonEvent: The newly generated non-thunk event.
```
almann: Same as above.

tests/test_writer_binary_raw.py
```diff
@@ -156,14 +171,19 @@
         (_D("1e-1"), [0x52, 0xC1, 0x01]),
         (_D("0e-1"), [0x51, 0xC1]),
         (_D("0e1"), [0x51, 0x81]),
+        (_D("-1e1"), [0x52, 0x81, 0x81]),
+        (_D("-1e0"), [0x52, 0x80, 0x81]),
+        (_D("-1e-1"), [0x52, 0xC1, 0x81]),
+        (_D("-0e-1"), [0x52, 0xC1, 0x80]),
+        (_D("-0e1"), [0x52, 0x81, 0x80]),
```
almann: Is there any reason to model the test in terms of list of int rather than byte string?

tests/writer_util.py
```diff
@@ -14,60 +14,49 @@
...
 
-_E = IonEvent
-_IT = IonType
-_ET = IonEventType
-
-_STREAM_END_EVENT = (_E(_ET.STREAM_END),)
-
-
-class _P(record('desc', 'events', 'expected')):
+class P(record('desc', 'events', 'expected')):
```
almann: I'm probably not being clear, but a name that is more than `P` is a good idea for test code that is to be *shared*.  For example, `WriterParameters`.  This can be idiomatically bound to `_P` or something in the individual test modules.

almann:
Looks good--minor comments above and I think we're basically good to merge.

amazon/ion/core.py
tgregg: Yep, I missed that. Fixed.

amazon/ion/core.py
tgregg: Fixed.

tests/test_writer_binary_raw.py
tgregg: At first I had trouble with python 2/3 compatibility since `bytes` semantics changed, but after a little more fiddling I got something working.

tests/writer_util.py
tgregg: Done.

tests/test_writer_binary_raw.py
almann: Take a look at [Six's binary/text utilities](https://pythonhosted.org/six/#binary-and-text-data), in particular `indexbytes`, `byte2int`, and `int2byte`.

tgregg:
I think this is ready for merge. Absent any objections, I'm going to merge into master at 4pm.

PR #24, opened 2016-07-12
Compatibility fixes for Python 2.6

### LINKED ISSUE #19: Fix Python 2.6 Compatibility
Python 2.6 is disabled in Travis CI, the principal reason seems to be about assumptions of float to string encoding and it being not as normalized as in later versions.

We need to fix the tests and re-enable the Travis CI config.

almann:
Fixes text writer test to work on Python 2.6.
Eliminates some redundant tests.
Adds bit_length(int) and total_seconds(timedelta) in util.
Adds Python 2.6 to Travis CI.

Fixes #19


tests/test_writer_binary_raw.py
```diff
@@ -152,16 +152,12 @@
...
-        (0e-0, b'\x40'),
         (1e0, b'\x48\x3F\xF0\x00\x00\x00\x00\x00\x00'),
-        (1e-0, b'\x48\x3F\xF0\x00\x00\x00\x00\x00\x00'),
         (-1e0, b'\x48\xBF\xF0\x00\x00\x00\x00\x00\x00'),
         (1e1, b'\x48\x40\x24\x00\x00\x00\x00\x00\x00'),
         (-1e1, b'\x48\xC0\x24\x00\x00\x00\x00\x00\x00'),
         (float('inf'), b'\x48\x7F\xF0\x00\x00\x00\x00\x00\x00'),
-        (1e400, b'\x48\x7F\xF0\x00\x00\x00\x00\x00\x00'),
         (float('-inf'), b'\x48\xFF\xF0\x00\x00\x00\x00\x00\x00'),
-        (-1e400, b'\x48\xFF\xF0\x00\x00\x00\x00\x00\x00'),
```
almann: These cases and those above it are technically redundant, the Python data model of this text is the same value so we aren't testing a different value.

tests/writer_util.py
```diff
@@ -36,7 +36,7 @@ def _scalar_p(ion_type, value, expected, force_stream_end):
     if force_stream_end:
         events += _STREAM_END_EVENT
     return WriterParameter(
-        desc='SCALAR %s - %s' % (ion_type.name, expected),
+        desc='SCALAR %s - %r - %r' % (ion_type.name, value, expected),
```
almann: This causes failure to render for raw byte sequences that cannot be encoded as UTF-8 making it impossible to get the actual failure from the report.

tgregg:
Looks good.

PR #31, opened 2016-09-08
simpleion API

tgregg:
The starting point for a simplejson-esque API which allows users to easily convert Python objects to and from Ion data (currently binary only).


almann:
Looks good, pushed as:

ed6db9bdd7b85261297291afd0cacf7a44fb61bb

PR #32, opened 2016-10-19
Ion type algorithm now inspects entire type hierarchy

mijamo911:
Previously the algorithm would fail given a subclass of any supported class


tests/test_simple_types.py
```diff
@@ -106,3 +107,9 @@ def test_event_types(p):
     assert value_output.ion_event is to_event_output
     assert value_output.ion_type is ion_type
     assert p.event.annotations == value_output.ion_annotations
+
+def test_subclass_types():
+    class Foo(dict):
+        pass
+
+    assert _ion_type(Foo()) is IonType.STRUCT
```
tgregg: No newline at end of file.

PR #39, opened 2017-01-21
Adds non-blocking text reader.

tgregg:
Includes:
-simpleion text support
-equivalence module for comparing Python objects under the Ion data model
-test harness for ion-tests vectors


tgregg:
Investigating why python2 succeeds all tests when run locally via tox.

tgregg:
Local build of python 2.7 was a narrow (UCS2) build, whereas the one used here is a wide build. Fixed a small bug that caused wide python 2 builds to fail on resurrection of non-BMP characters. This should resolve the 2.7 and pypy failures.

Investigating the 2.6 failures now.

tgregg:
Python 2.6 doesn't support bytearray as input to base64.b64decode. Quick fix.

Issue #37, opened 2017-01-21
Fix validation of binary annotation wrapper length subfields


tgregg:
The binary reader needs improved validation of annotation wrappers' various length subfields.

The following is from `bad/container_length_mismatch.10n`:

`\xe0\x01\x00\xea\xe6\x81\x84\x71\x04\x71\x04`

This contains a version marker, followed by an annotation wrapper of total specified length 6 octets (type descriptor \xe6), followed by a one-octet annot_length field (\x81) which specifies that the annotation SIDs (VarUInts) total one octet, followed by octet \x84 which corresponds to SID 4, followed by two symbol values representing SID 4. This passes, returning two unannotated symbol 4 values. However, it should fail because the annotation wrapper's value ended before reaching the 6-octet length specified in the type descriptor.

Annotation wrappers are invalid unless their type descriptor length == (the length of the annot_length subfield) + (the UInt value of the annot_length subfield) + (the total length of the wrapped value (the first value following the annot subfield(s))). In this case, 6 != (1 + 1 + 2).

Similarly, annotation wrappers with no annot subfields should be rejected. The following is from `bad/emptyAnnotatedInt.10n`:
`\xe0\x01\x00\xea\xe3\x80\x21\x01`

This contains a version marker, followed by an annotation wrapper of total specified length 3 octets (type descriptor \xe3), followed by a one-octed annot_length field (\x80) which specifies that the annotation SIDs total zero octets (i.e. there aren't any annotation SIDs), followed by the binary representation of the int value 1. This passes, returning an unannotated int 1. However, it should fail, as annotation wrappers must have at least one annotation SID.

Successful resolution of this issue will involve removal of the following files from the `test_vectors.py` skip list:
`bad/container_length_mismatch.10n`
`bad/emptyAnnotatedInt.10n`

tgregg:
Out for PR in #44 .

PR #44, opened 2017-01-25
Validates total length when reading binary annotation wrapper subfields.

tgregg:
See #37 .

In short, the length declared by the annotation wrapper must exactly match the number of octets taken up by its subfields (including its value). This adds that validation logic.

Note: test vectors to exercise these cases have been added to ion-tests [here](https://github.com/amznlabs/ion-tests/pull/23).


tgregg:

Issue #38, opened 2017-01-21
Fix binary read support for ordered structs


tgregg:
According to the spec for the `struct` type, there is a special case for the type descriptor \xD1: "When L is 1, the struct has at least one symbol/value pair, the length field exists, and the field name integers are sorted in increasing order."

The binary reader currently doesn't support this case, failing on structs defined this way for overrunning one byte. Support for the special case of a struct with L=1 should be added.

Successful resolution of this issue will involve removal of the following files from the `test_vectors.py` skip list:
`good/structAnnotatedOrdered.10n`
`good/structOrdered.10n`

tgregg:
Out for PR in #45 .

PR #45, opened 2017-01-26
Adds support for reading binary ordered structs (see #38).

tgregg:
See #38 .

In short, previously, the special case of type descriptor `0xd1` was not handled, resulting in length overrun for correctly formed ordered structs. This fixes that problem.

Note that the spec requires ordered structs to contain at least one field name/value pair. If that weren't required, there would be no need for `_ordered_struct_start_handler`. There is an ion-tests [pull request](https://github.com/amznlabs/ion-tests/pull/24) which adds a bad case to exercise this.


tgregg:

PR #47, opened 2017-01-31
Add roundtrips to test_vectors; fix bugs uncovered during the process.

tgregg:
Bugs fixed:
* The text reader now only considers fractional precision valid if seconds precision was found.
* If the binary writer is given a `SymbolToken` with non-`None` text, it will now perform its LST lookup using that text instead of the SID, if any. Previously, it would look up by SID and error if the input `SymbolToken`'s text was different than the mapping already in the LST (if any). It's valid for SIDs to be reassigned for symbols with known text.
* The binary writer now preserves negative-zero floats by only using the one-octet float zero representation when the input is *positive* zero.
* The binary writer's logic for computing fractional precision is fixed for the zero-coefficient case.
* The raw binary field writer now uses the correct mask when writing the middle octets of variable length fields of size 3 octets or larger. Additional test coverage has been added.
* The text writer's timestamp serialization logic has been rewritten to take precision into account (this was previously left as a TODO).
* The text writer now escapes backslash characters in quoted text.

Additions to `test_vectors.py`:
* Roundtrip through text and binary for all `good/` vectors. Verify equality of the objects before and after roundtripping.
* For each `equivs`/`nonequivs` pair, test eight additional combinations (where `equivs` pairs must still be equivalent after the roundtrip and `nonequivs` pairs must still be non-equivalent):
    * A control, B binary roundtrip
    * A control, B text roundtrip
    * A binary roundtrip, B control
    * A text roundtrip, B control
    * A binary roundtrip, B binary roundtrip
    * A text roundtrip, B binary roundtrip
    * A binary roundtrip, B text roundtrip
    * A text roundtrip, B text roundtrip.
* Note that these changes are dependent on a few small tweaks to ion-tests. See the [pull request](https://github.com/amznlabs/ion-tests/pull/25) for those changes.


tgregg:

PR #50, opened 2017-05-09
Bumps version to 0.1.1; updates URLs

tgregg:
* The last version on PyPI is from Oct 20, 2016, and doesn't include text reading support. I plan to release this version as 0.1.1.
* URLs have been updated from amznlabs -> amzn


pbcornell:
lgtm!

PR #56, opened 2017-10-08
Adding an implementation of simpleion.dumps

Antrikshy:
I implemented `simpleion.dumps` as a simple wrapper over `simpleion.dump` using `six.BytesIO` as the file buffer.

I got an error when running existing tests:

    tests/test_vectors.py:328: in <module>
        _basic_params(_T.BAD, _BAD_UTF8_SUBDIR),
    tests/test_vectors.py:199: in _basic_params
        for file in _list_files(directory_path):
    tests/test_vectors.py:174: in _list_files
        for file in listdir(directory_path):
    E   OSError: [Errno 2] No such file or directory: '/path/to/ion-python/vectors/iontestdata/good'

Does it have to do with my change or something I'm missing in my dev environment?

I hope contributions are accepted!

PR #57, opened 2017-10-22
Adding an implementation of simpleion.loads

### LINKED ISSUE #42: Implement dumps and loads in simpleion module.
Now that `dump` and `load` are implemented with text and binary support, this should be simple.

Antrikshy:
Now that #56 is merged, this resolves #42.

I was looking for a weekend project, figured why not do this and actually go for that Hacktoberfest t-shirt! :)

Let me know if you have a better name for `_assert_symbol_aware_ion_equals`.

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -21,7 +21,7 @@ from __future__ import print_function
 
 from datetime import datetime
 from decimal import Decimal
-from io import TextIOBase
+from io import BytesIO, TextIOBase
 from itertools import chain
 
 import six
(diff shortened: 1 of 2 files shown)


amazon/ion/simpleion.py
```diff
@@ -372,7 +372,35 @@ def add(obj):
...
+        use_decimal: NOT IMPLEMENTED
+        **kw: NOT IMPLEMENTED
+
+    Returns (Any):
+        if single_value is True:
+            A Python object representing a single Ion value.
+        else:
+            A sequence of Python objects representing a stream of Ion values.
+    """
+    ion_buffer = BytesIO(ion_str.encode())
```
tgregg: Does this work for both binary and text types? This API should be able to work for both; users should be able to use this to read both text and binary Ion data, e.g.

`res = loads(b'\xE0\x01\x00\xEA\x20')`

and

`res = loads(u'$ion_1_0 0')`

Symmetrically, perhaps `dumps` should work for both too. The tricky part is keeping consistent with `simplejson.dumps`, which looks like its return type changed from a binary type to a text type from Python 2 -> 3 (because it remained `str` the whole time). Maybe we can just return a binary type for both text and binary Ion? Or return a text type for text Ion? Thoughts?

Antrikshy: Huh, I didn't know simplejson had implicit behavior like that. Seems kinda weird, but then I've never used the library myself. Tracing through the code, it has [this decoder class](https://github.com/simplejson/simplejson/blob/master/simplejson/decoder.py#L363-L400) thing going on. I can't tell where the hex conversion happens though.

Is a byte input expected to be hex encoded like this? How does the `encoding` parameter play into this, if at all?

almann:
@Antrikshy, sorry for the long delay here--I was looking to use `loads` and `dumps` and realized that we've left this PR languishing.  I've pulled in your PR and adapted it to work with binary as #62.

Issue #74, opened 2019-03-01
Tests fail on i586 for Python 3.7


jayvdb:
```console
[  254s] ==================================== ERRORS ====================================
[  254s] ____________________ ERROR collecting tests/test_vectors.py ____________________
[  254s] tests/test_vectors.py:72: in <module>
[  254s]     getcontext().Emax = 100000000000000000
[  254s] E   OverflowError: Python int too large to convert to C ssize_t
[  254s] !!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
[  254s] =========================== 1 error in 13.40 seconds ===========================
```

https://build.opensuse.org/package/live_build_log/home:jayvdb:moban/python-amazon.ion/openSUSE_Tumbleweed/i586

tgregg:
#75 attempts to fix this particular issue. I don't have access to this platform, though, so I could not verify or check for additional platform-specific errors.

tgregg:
@jayvdb please reopen this issue if you encounter more errors related to i586.

jayvdb:
Now I get a slightly different error:

```console
[  244s] ____________________ ERROR collecting tests/test_vectors.py ____________________
[  244s] tests/test_vectors.py:73: in <module>
[  244s]     getcontext().Emax = 2147483647
[  244s] E   ValueError: valid range for Emax is [0, MAX_EMAX]
```

https://build.opensuse.org/package/show/home:jayvdb:moban/python-amazon.ion

I cant re-open; insufficient rights ;-)

tgregg:
I didn't know about the MAX_EMAX and MAX_EMIN constants; looks like they were [introduced](https://python.readthedocs.io/en/stable/whatsnew/3.3.html#id2) in 3.3. The values for a 32-bit architecture are
MAX_EMAX 425000000
MIN_EMIN -425000000

I'm not sure why these limits are less than the max size of an `ssize_t`, which seems to be the type they're stored in in C. Unfortunately, a few of our tests require values outside this range, meaning we need an architecture-specific skip list.

PR #77, opened 2019-03-04
Lowers the configured Emax and Emin in test_vectors into the platform-specific ranges.

tgregg:
*Issue #, if available:*
#74 

*Description of changes:*
The decimal Emax and Emin limits are lower on 32-bit systems. They are even lower than the range of an `ssize_t`. Because the `subfieldVarInt.ion` test contains a decimal with a negative exponent that is less than the 32-bit Emin, this test must be skipped on 32-bit platforms.

### DIFF
--- a/tests/test_vectors.py
+++ b/tests/test_vectors.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import struct
 from collections import defaultdict
 from decimal import getcontext
 from functools import partial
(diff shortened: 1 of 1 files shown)


tests/test_vectors.py
```diff
@@ -67,11 +68,18 @@
...
+
 # Set these Decimal limits arbitrarily high because some test vectors require it. If users need decimals this large,
-# they'll have to do this in their code too. CPython may store these in an ssize_t, so the values should remain in the
-# 32-bit range for compatibility with 32-bit platforms.
-getcontext().Emax = 2147483647
-getcontext().Emin = -2147483648
+# they'll have to do this in their code too. Note: these values are equivalent to the MAX_EMAX and MAX_EMIN values
+# for the platform's architecture (32 or 64 bit); the constants were introduced in Python 3.3.
+if _PLATFORM_ARCHITECTURE == 32:
+    getcontext().Emax = 425000000
```
zslayton: Is there a way we can reference the constants directly instead of reproducing the numeric literals? If not, could you add a link to [the newly defined values](https://python.readthedocs.io/en/stable/whatsnew/3.3.html#id2) in the comment?

jayvdb:
fwiw, this patch (https://github.com/amzn/ion-python/pull/77/commits/83a51a0fd7122fd507cb9595837e9135afc63db7) did fix the i586 builds on openSUSE
https://build.opensuse.org/package/show/home:jayvdb:moban/python-amazon.ion

Issue #35, opened 2017-01-20
Support text timestamps with more than 6 digits of fractional precision


tgregg:
The text reader does not support timestamps with greater than 6 digits fractional precision because Python's datetime only supports microsecond precision. Current logic raises an error instead of silently rounding/truncating.

Consider adding an extension to ion-python's `Timestamp` type to allow roundtripping of arbitrary fractional precision with full fidelity.

Successful resolution of this issue will involve removal of the following file from the `test_vectors.py` skip list:
`good/equivs/timestampsLargeFractionalPrecision.ion`

zslayton:
This appears to have been closed by #100, which [removed `good/equivs/timestampsLargeFractionalPrecision.ion` from the skip list](https://github.com/amzn/ion-python/pull/100/files#diff-60f006a04937736c49b7592e4a7388f2L99).

PR #100, opened 2019-09-17
Support timestamps with more than 6 digits of fractional precision

avneet-toor-bq:
Issue #35 

- Store the fractional seconds component as a property on the Timestamp object, storing full fidelity.
- Truncate the fractional seconds to microseconds, if need be, and pass that to the Timestamp/datetime.
- When serializing said Timestamp object the writer will use that fractional seconds field to encode the timestamp in text/binary with whatever fidelity it was set. 

### DIFF
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -108,15 +108,12 @@ def scale_to_precision(fractional_seconds, fractional_precision):
     """Scales the ```fractional_seconds``` attribute by -```fractional_precision```.
 
     Args:
-        fractional_seconds (int): The numerical value to scale.
+        fractional_seconds (Decimal): The numerical value to scale.
         fractional_precision (int): The negative of this numerical value to scale ```fractional_seconds``` by.
     Returns:
         Decimal: The scaled value.
     """
-    fractional_seconds = Decimal(str(fractional_seconds)[:fractional_precision])
-    for i in range(fractional_precision):
-        fractional_seconds = Decimal(fractional_seconds) / 10
-    return fractional_seconds
+    return Decimal(fractional_seconds) * Decimal(Decimal(10) ** Decimal(-fractional_precision))
 
 
 class IonEvent(record(
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -124,7 +124,7 @@ def _fractional_seconds_to_microseconds(fractional_seconds):
     Returns:
         int: The scaled value.
     """
-    integral_fractional_seconds = fractional_seconds * (10 ** MICROSECOND_PRECISION)
+    integral_fractional_seconds = fractional_seconds * BASE_TEN_MICROSECOND_PRECISION_EXPONENTIATION
     return int(integral_fractional_seconds)
 
 
@@ -400,6 +400,7 @@ TIMESTAMP_PRECISION_FIELD = 'precision'
 TIMESTAMP_FRACTION_PRECISION_FIELD = 'fractional_precision'
 TIMESTAMP_FRACTIONAL_SECONDS_FIELD = 'fractional_seconds'
 MICROSECOND_PRECISION = 6
+BASE_TEN_MICROSECOND_PRECISION_EXPONENTIATION = 10 ** MICROSECOND_PRECISION
 
 
 class Timestamp(datetime):
@@ -480,6 +481,9 @@ class Timestamp(datetime):
         if fractional_seconds is not None and (fractional_precision is not None or datetime_microseconds is not None):
             raise ValueError('fractional_seconds cannot be specified '
                              'when fractional_precision or microseconds are not None.')
+
+        if fractional_precision is not None and datetime_microseconds is None:
+            raise ValueError('datetime_microseconds cannot be None while fractional_precision is not None.')
 
         if fractional_seconds is not None:
             fractional_precision = min(fractional_seconds.as_tuple().exponent * -1, MICROSECOND_PRECISION)
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -109,10 +109,11 @@ def scale_to_precision(fractional_seconds, fractional_precision):
 
     Args:
         fractional_seconds (int): The numerical value to scale.
-        fractional_precision (int): The numerical value to scale ```fractional_seconds``` by.
+        fractional_precision (int): The negative of this numerical value to scale ```fractional_seconds``` by.
     Returns:
         Decimal: The scaled value.
     """
+    fractional_seconds = Decimal(str(fractional_seconds)[:fractional_precision])
     for i in range(fractional_precision):
         fractional_seconds = Decimal(fractional_seconds) / 10
     return fractional_seconds
@@ -412,7 +413,7 @@ class Timestamp(datetime):
             - If microsecond/fractional_precision is specified but fractional_seconds is not, a new Decimal is
             created from the value of microsecond and truncated to fractional_precision.
 
-            - If fractional_seconds is specified but microsecond/fractional_precision is not, microsecond is set
+            - If fractional_seconds is specified but microsecond is not, microsecond is set
             to the value of fractional_seconds scaled up to microseconds magnitude and fractional_precision is set to
             the minimum of the precision of fractional_seconds and MICROSECOND_PRECISION.
 
@@ -425,65 +426,55 @@ class Timestamp(datetime):
         precision = None
         fractional_precision = None
         fractional_seconds = None
-        fractional_seconds2 = None
+        print(len(args))
         MICROSECOND_ARGUMENT_INDEX = 6
+        DATETIME_MICROSECONDS = None
+        if len(args) > 6:
+            DATETIME_MICROSECONDS = args[MICROSECOND_ARGUMENT_INDEX]
         if TIMESTAMP_PRECISION_FIELD in kwargs:
             precision = kwargs.get(TIMESTAMP_PRECISION_FIELD)
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_PRECISION_FIELD]
         if TIMESTAMP_FRACTION_PRECISION_FIELD in kwargs:
             fractional_precision = kwargs.get(TIMESTAMP_FRACTION_PRECISION_FIELD)
-            # print(fractional_precision)
             if fractional_precision is not None and fractional_precision < 1:
-                fractional_precision = 1
-                # raise ValueError('Cannot construct a Timestamp with fractional precision of %d digits, '
-                #                  'which is less than 1.' % fractional_precision)
+                raise ValueError('Cannot construct a Timestamp with fractional precision of %d digits, '
+                                 'which is less than 1.' % fractional_precision)
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTION_PRECISION_FIELD]
         if TIMESTAMP_FRACTIONAL_SECONDS_FIELD in kwargs:
             fractional_seconds = kwargs.get(TIMESTAMP_FRACTIONAL_SECONDS_FIELD)
-            fractional_seconds2 = fractional_seconds
             # Fractional seconds should be a Decimal in the range [0, 1), or None.
             if fractional_seconds is not None and fractional_seconds > 1:
                 raise ValueError('Fractional seconds cannot be greater than 1.')
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTIONAL_SECONDS_FIELD]
+        if len(args) > 6:
+            if DATETIME_MICROSECONDS is not None and fractional_precision is None:
+                raise ValueError('Fractional precision cannot be None while microsecond is not None.')
 
-        # if args[MICROSECOND_ARGUMENT_INDEX] is not None and fractional_precision is None:
-        #     raise ValueError('Fractional precision cannot be None while microsecond is not None.')
+            if DATETIME_MICROSECONDS is not None and fractional_precision is not None \
+                    and fractional_seconds is not None:
+                if scale_to_precision(DATETIME_MICROSECONDS, fractional_precision) != fractional_seconds:
+                    raise ValueError('Microseconds and fractional seconds are not equivalent.')
 
-        if args[MICROSECOND_ARGUMENT_INDEX] is not None and fractional_precision is not None \
-                and fractional_seconds is not None:
-            if scale_to_precision(args[MICROSECOND_ARGUMENT_INDEX], fractional_precision) != fractional_seconds:
-                raise ValueError('Microseconds and fractional seconds are not equivalent.')
+            if DATETIME_MICROSECONDS is not None and fractional_precision is not None \
+                    and fractional_seconds is None:
+                fractional_seconds = scale_to_precision(DATETIME_MICROSECONDS, MICROSECOND_PRECISION)
 
-        if args[MICROSECOND_ARGUMENT_INDEX] is not None and fractional_precision is not None \
-                and fractional_seconds is None:
-            ss = args[MICROSECOND_ARGUMENT_INDEX]
-            fractional_seconds = scale_to_precision(args[MICROSECOND_ARGUMENT_INDEX], MICROSECOND_PRECISION)
-            # fractional_seconds2 = fractional_seconds
+            if DATETIME_MICROSECONDS is None and fractional_seconds is not None:
+                lst = list(args)
+                integral_fractional_seconds = fractional_seconds * (10**MICROSECOND_PRECISION)
+                lst[MICROSECOND_ARGUMENT_INDEX] = Decimal(integral_fractional_seconds).quantize(MICROSECOND_PRECISION,
+                                                                                                rounding="ROUND_DOWN")
+                args = tuple(lst)
+                DATETIME_MICROSECONDS = args[MICROSECOND_ARGUMENT_INDEX]
 
-        if args[MICROSECOND_ARGUMENT_INDEX] is None and fractional_seconds is not None:
-            # if fractional_seconds != 0:
-            #     fraction_seconds_length = len(str(fractional_seconds)) - str(fractional_seconds).index('.') - 1
-            # else:
-            #     fraction_seconds_length = fractional_seconds
-            # fractional_precision = min(fraction_seconds_length, MICROSECOND_PRECISION)
-            lst = list(args)
-            integral_fractional_seconds = fractional_seconds * (10**MICROSECOND_PRECISION)
-            lst[MICROSECOND_ARGUMENT_INDEX] = Decimal(integral_fractional_seconds).quantize(MICROSECOND_PRECISION,
-                                                                                            rounding="ROUND_DOWN")
-            # lst[MICROSECOND_ARGUMENT_INDEX] = int(str(integral_fractional_seconds)[:MICROSECOND_PRECISION])
-            # lst[MICROSECOND_ARGUMENT_INDEX] = int(str(lst[MICROSECOND_ARGUMENT_INDEX])[0:MICROSECOND_PRECISION])
-            args = tuple(lst)
-            print(args)
-
-        # If microsecond is still None, set it to 0.
-        elif args[MICROSECOND_ARGUMENT_INDEX] is None:
-            lst = list(args)
-            lst[MICROSECOND_ARGUMENT_INDEX] = 0
-            fractional_precision = 989898
-            args = tuple(lst)
+            # If microsecond is still None, set it to 0.
+            if DATETIME_MICROSECONDS is None:
+                lst = list(args)
+                lst[MICROSECOND_ARGUMENT_INDEX] = 0
+                args = tuple(lst)
 
         instance = super(Timestamp, cls).__new__(cls, *args, **kwargs)
         setattr(instance, TIMESTAMP_PRECISION_FIELD, precision)
@@ -503,11 +494,8 @@ class Timestamp(datetime):
     def adjust_from_utc_fields(*args, **kwargs):
         """Constructs a timestamp from UTC fields adjusted to the local offset if given."""
         raw_ts = Timestamp(*args, **kwargs)
-        # print(raw_ts.fractional_precision)
         offset = raw_ts.utcoffset()
         if offset is None or offset == timedelta():
-            # print(args)
-            # print(kwargs)
             return raw_ts
 
         # XXX This returns a datetime, not a Timestamp (which has our precision if defined)
@@ -523,7 +511,7 @@ class Timestamp(datetime):
             adjusted.hour,
             adjusted.minute,
             adjusted.second,
-            adjusted.microsecond,
+            None,
             raw_ts.tzinfo,
             precision=raw_ts.precision,
             fractional_precision=raw_ts.fractional_precision,
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -434,6 +434,10 @@ class Timestamp(datetime):
             args = tuple(lst)
             fractional_precision = min(fraction_seconds_length, MICROSECOND_PRECISION)
 
+        if args[6] is None:
+            lst = list(args)
+            lst[6] = 0
+            args = tuple(lst)
         instance = super(Timestamp, cls).__new__(cls, *args, **kwargs)
         setattr(instance, TIMESTAMP_PRECISION_FIELD, precision)
         setattr(instance, TIMESTAMP_FRACTION_PRECISION_FIELD, fractional_precision)
@@ -507,7 +511,6 @@ def timestamp(year, month=1, day=1,
         if fractional_precision is None:
             fractional_precision = MICROSECOND_PRECISION
     else:
-        microsecond = 0
         if fractional_precision is not None:
             raise ValueError('Cannot have fractional precision without a fractional component.')
     return Timestamp(
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -104,20 +104,20 @@ class IonEventType(Enum):
         return self < 0
 
 
-def scale_to_precision(fractional_seconds, fractional_precision):
+def _scale_to_precision(microseconds, fractional_precision):
     """Scales the ```fractional_seconds``` attribute by -```fractional_precision```.
 
     Args:
-        fractional_seconds (Decimal): The numerical value to scale.
+        microseconds (int): The numerical value to scale.
         fractional_precision (int): The negative of this numerical value to scale ```fractional_seconds``` by.
     Returns:
         Decimal: The scaled value.
     """
-    return Decimal(fractional_seconds).scaleb(-fractional_precision)
+    return Decimal(microseconds).scaleb(-fractional_precision)
 
 
-def round_to_sixth_precision(fractional_seconds):
-    """Scales the ```fractional_seconds``` attribute to a six digit integer.
+def _fractional_seconds_to_microseconds(fractional_seconds):
+    """Scales the ```fractional_seconds``` attribute to microseconds.
 
     Args:
         fractional_seconds (Decimal): The numerical value to scale.
@@ -184,12 +184,12 @@ class IonEvent(record(
                 if (self_fractional_seconds is None and other_fractional_seconds is not None) and \
                         (other_fractional_precision != self_fractional_precision
                          or other_fractional_seconds !=
-                         scale_to_precision(self.value.microsecond, self_fractional_precision)):
+                         _scale_to_precision(self.value.microsecond, self_fractional_precision)):
                     return False
                 if (self_fractional_seconds is not None and other_fractional_seconds is None) and \
                         (self_fractional_precision != other_fractional_precision
                          or self_fractional_seconds !=
-                         scale_to_precision(other.value.microsecond, other_fractional_precision)):
+                         _scale_to_precision(other.value.microsecond, other_fractional_precision)):
                     return False
 
             if isinstance(self.value, datetime):
@@ -446,53 +446,48 @@ class Timestamp(datetime):
             del kwargs[TIMESTAMP_PRECISION_FIELD]
         if TIMESTAMP_FRACTION_PRECISION_FIELD in kwargs:
             fractional_precision = kwargs.get(TIMESTAMP_FRACTION_PRECISION_FIELD)
-            if fractional_precision is not None and fractional_precision < 1:
+            if fractional_precision is not None and 1 > fractional_precision > MICROSECOND_PRECISION:
                 raise ValueError('Cannot construct a Timestamp with fractional precision of %d digits, '
-                                 'which is less than 1.' % fractional_precision)
+                                 'which is out of the supported range of [1, %d].'
+                                 % (fractional_precision, MICROSECOND_PRECISION))
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTION_PRECISION_FIELD]
         if TIMESTAMP_FRACTIONAL_SECONDS_FIELD in kwargs:
             fractional_seconds = kwargs.get(TIMESTAMP_FRACTIONAL_SECONDS_FIELD)
-            # Fractional seconds should be a Decimal in the range [0, 1), or None.
-            if fractional_seconds is not None and fractional_seconds > 1:
-                raise ValueError('Fractional seconds cannot be greater than 1.')
+            if fractional_seconds is not None:
+                fractional_seconds = Decimal(fractional_seconds)
+                if 0 > fractional_seconds >= 1:
+                    raise ValueError('Cannot construct a Timestamp with fractional seconds of %s, '
+                                     'which is out of the supported range of [0, 1).'
+                                     % str(fractional_seconds))
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTIONAL_SECONDS_FIELD]
-        if len(args) > 6:
-            if datetime_microseconds is not None and fractional_precision is not None \
-                    and fractional_seconds is not None:
-                if scale_to_precision(datetime_microseconds, fractional_precision) != fractional_seconds:
-                    raise ValueError('Microseconds and fractional seconds are not equivalent.')
 
-            if fractional_seconds is not None and fractional_precision is None:
-                raise ValueError('Fractional precision cannot be None while fractional seconds is not None.')
+        if fractional_seconds is not None and (fractional_precision is not None or datetime_microseconds is not None):
+            raise ValueError('fractional_seconds cannot be specified '
+                             'when fractional_precision or microseconds are not None.')
 
-            if datetime_microseconds is not None and fractional_precision is not None \
-                    and fractional_seconds is None:
-                if fractional_precision > 6:
-                    fractional_seconds = scale_to_precision(datetime_microseconds, fractional_precision)
-                    lst = list(args)
-                    lst[microsecond_argument_index] = round_to_sixth_precision(fractional_seconds)
-                    args = tuple(lst)
-                else:
-                    fractional_seconds = scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION)
+        if fractional_precision is not None and datetime_microseconds is None:
+            raise ValueError('Cannot have fractional precision without a fractional component.')
 
-            if datetime_microseconds is None and fractional_seconds is not None:
-                lst = list(args)
-                lst[microsecond_argument_index] = round_to_sixth_precision(fractional_seconds)
-                args = tuple(lst)
-                datetime_microseconds = args[microsecond_argument_index]
-
-            # If microsecond is still None, set it to 0.
-            if datetime_microseconds is None:
+        if (datetime_microseconds is None) \
+                and fractional_precision is None and fractional_seconds is None:
+            if len(args) > 6:
                 lst = list(args)
                 lst[microsecond_argument_index] = 0
                 args = tuple(lst)
-
-        if datetime_microseconds is not None and datetime_microseconds > 999999:
-            lst = list(args)
-            lst[microsecond_argument_index] = int(str(datetime_microseconds)[:MICROSECOND_PRECISION])
-            args = tuple(lst)
+        else:
+            if datetime_microseconds is not None:
+                if fractional_precision is None:
+                    fractional_precision = MICROSECOND_PRECISION
+                fractional_seconds = str(_scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION))
+                fractional_seconds = Decimal(fractional_seconds[0:(fractional_precision + 2)])
+            elif fractional_seconds is not None:
+                fractional_precision = min(fractional_seconds.as_tuple().exponent * -1, MICROSECOND_PRECISION)
+                datetime_microseconds = _fractional_seconds_to_microseconds(fractional_seconds)
+                lst = list(args)
+                lst[microsecond_argument_index] = datetime_microseconds
+                args = tuple(lst)
 
         instance = super(Timestamp, cls).__new__(cls, *args, **kwargs)
         setattr(instance, TIMESTAMP_PRECISION_FIELD, precision)
@@ -532,7 +527,7 @@ class Timestamp(datetime):
             None,
             raw_ts.tzinfo,
             precision=raw_ts.precision,
-            fractional_precision=raw_ts.fractional_precision,
+            fractional_precision=None,
             fractional_seconds=raw_ts.fractional_seconds
         )
 
@@ -564,12 +559,6 @@ def timestamp(year, month=1, day=1,
     if delta is not None:
         tz = OffsetTZInfo(delta)
 
-    if microsecond is not None or fractional_seconds is not None:
-        if fractional_precision is None:
-            fractional_precision = MICROSECOND_PRECISION
-    else:
-        if fractional_precision is not None:
-            raise ValueError('Cannot have fractional precision without a fractional component.')
     return Timestamp(
         year, month, day,
         hour, minute, second, microsecond,
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -21,7 +21,7 @@ from __future__ import print_function
 
 from collections import MutableMapping, MutableSequence
 from datetime import datetime, timedelta, tzinfo
-from decimal import Decimal
+from decimal import Decimal, ROUND_FLOOR
 from math import isnan
 
 import six
@@ -104,16 +104,31 @@ class IonEventType(Enum):
         return self < 0
 
 
-def _scale_to_precision(microseconds, fractional_precision):
-    """Scales the ```fractional_seconds``` attribute by -```fractional_precision```.
+def _scale_to_precision(value, fractional_precision):
+    """Scales the ```value``` attribute by -```fractional_precision```.
+
+    If the length of the value is greater than the ```fractional_precision```, it is determined if the extra digits in
+    ```value``` are not significant, i.e. have the value '0'.
+    If a significant digit is found, a ValueError is raised.
+    If no significant digits are not found, the ```value``` is truncated to match that of the```fractional_precision```
+    and the scaling is performed.
 
     Args:
-        microseconds (int): The numerical value to scale.
+        value (int): The numerical value to scale.
         fractional_precision (int): The negative of this numerical value to scale ```fractional_seconds``` by.
     Returns:
         Decimal: The scaled value.
     """
-    return Decimal(microseconds).scaleb(-fractional_precision)
+    value_string = str(value)
+    if len(value_string) > fractional_precision:
+        for digit in value_string[fractional_precision:]:
+            if digit != '0':
+                raise ValueError('Found timestamp fractional precision of %d digits, '
+                                 'which is less than needed to serialize %d microseconds.'
+                                 % (fractional_precision, value))
+        value_string = value_string[0:fractional_precision]
+        value = int(value_string)
+    return Decimal(value).scaleb(-fractional_precision)
 
 
 def _fractional_seconds_to_microseconds(fractional_seconds):
@@ -401,6 +416,14 @@ TIMESTAMP_FRACTION_PRECISION_FIELD = 'fractional_precision'
 TIMESTAMP_FRACTIONAL_SECONDS_FIELD = 'fractional_seconds'
 MICROSECOND_PRECISION = 6
 BASE_TEN_MICROSECOND_PRECISION_EXPONENTIATION = 10 ** MICROSECOND_PRECISION
+PRECISION_LIMIT_LOOKUP = {
+                            1: Decimal('0.' + '0' * (1 - 1) + '1'),
+                            2: Decimal('0.' + '0' * (2 - 1) + '1'),
+                            3: Decimal('0.' + '0' * (3 - 1) + '1'),
+                            4: Decimal('0.' + '0' * (4 - 1) + '1'),
+                            5: Decimal('0.' + '0' * (5 - 1) + '1'),
+                            6: Decimal('0.' + '0' * (6 - 1) + '1')
+                        }
 
 
 class Timestamp(datetime):
@@ -415,7 +438,7 @@ class Timestamp(datetime):
             * The ``fractional_precision`` field is passed as a keyword argument of the same name.
               This field only relates to to the ``microseconds`` field and can be thought of
               as the number of decimal digits that are significant.  This is an integer that
-              that is in the closed interval ``[0, 6]``.  If ``None``, ``microseconds`` must be
+              that is in the closed interval ``[0, 6]``.  If ``0``, ``microseconds`` must be
               ``0`` indicating no precision below seconds.  This argument is optional and only valid
               when ``microseconds`` is not ``None``.  If the ``microseconds`` specified has more
               precision than this field indicates, then that is an error.
@@ -435,7 +458,7 @@ class Timestamp(datetime):
             Consider some examples:
 
             * `2019-10-01T12:45:01Z` would have the following fields set:
-              * ``microseconds == 0``, ``fractional_precision == None``, ``fractional_seconds == None``
+              * ``microseconds == 0``, ``fractional_precision == 0``, ``fractional_seconds == Decimal('0')``
             * `2019-10-01T12:45:01.100Z` would have the following fields set:
               * ``microseconds == 100000``, ``fractional_precision == 3``, ``fractional_seconds == Decimal('0.100')``
             * `2019-10-01T12:45:01.123456789Z` would have the following fields set:
@@ -470,7 +493,6 @@ class Timestamp(datetime):
         if TIMESTAMP_FRACTIONAL_SECONDS_FIELD in kwargs:
             fractional_seconds = kwargs.get(TIMESTAMP_FRACTIONAL_SECONDS_FIELD)
             if fractional_seconds is not None:
-                fractional_seconds = Decimal(fractional_seconds)
                 if 0 > fractional_seconds >= 1:
                     raise ValueError('Cannot construct a Timestamp with fractional seconds of %s, '
                                      'which is out of the supported range of [0, 1).'
@@ -485,7 +507,12 @@ class Timestamp(datetime):
         if fractional_precision is not None and datetime_microseconds is None:
             raise ValueError('datetime_microseconds cannot be None while fractional_precision is not None.')
 
-        if fractional_seconds is not None:
+        if fractional_precision is not None and fractional_precision == 0 and datetime_microseconds != 0:
+            raise ValueError('datetime_microseconds cannot be non-zero while fractional_precision is 0.')
+
+        if fractional_precision is not None and fractional_precision == 0 and datetime_microseconds == 0:
+            fractional_seconds = Decimal('0e0')
+        elif fractional_seconds is not None:
             fractional_precision = min(fractional_seconds.as_tuple().exponent * -1, MICROSECOND_PRECISION)
             datetime_microseconds = _fractional_seconds_to_microseconds(fractional_seconds)
             lst = list(args)
@@ -499,8 +526,8 @@ class Timestamp(datetime):
                     lst[microsecond_argument_index] = datetime_microseconds
                     args = tuple(lst)
             else:
-                fractional_seconds = str(_scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION))
-                fractional_seconds = Decimal(fractional_seconds[0:(fractional_precision + 2)])
+                fractional_seconds = _scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION)\
+                    .quantize(PRECISION_LIMIT_LOOKUP[fractional_precision], rounding=ROUND_FLOOR)
 
         instance = super(Timestamp, cls).__new__(cls, *args, **kwargs)
         setattr(instance, TIMESTAMP_PRECISION_FIELD, precision)
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -169,27 +169,16 @@ class IonEvent(record(
                 # Special case for timestamps to capture equivalence over precision.
                 self_precision = getattr(self.value, TIMESTAMP_PRECISION_FIELD, None)
                 other_precision = getattr(other.value, TIMESTAMP_PRECISION_FIELD, None)
-                self_fractional_seconds = getattr(self.value, TIMESTAMP_FRACTIONAL_SECONDS_FIELD, None)
-                other_fractional_seconds = getattr(other.value, TIMESTAMP_FRACTIONAL_SECONDS_FIELD, None)
-                self_fractional_precision = getattr(self.value, TIMESTAMP_FRACTION_PRECISION_FIELD, None)
-                other_fractional_precision = getattr(other.value, TIMESTAMP_FRACTION_PRECISION_FIELD, None)
                 if self_precision != other_precision \
                         and not ((self_precision is None and other_precision is TimestampPrecision.SECOND) or
                                  (self_precision is TimestampPrecision.SECOND and other_precision is None)):
                     # The absence of precision indicates a naive datetime, which always has SECOND precision.
                     return False
-                if self_fractional_seconds is not None and other_fractional_seconds is not None \
-                        and self_fractional_seconds != other_fractional_seconds:
-                    return False
-                if (self_fractional_seconds is None and other_fractional_seconds is not None) and \
-                        (other_fractional_precision != self_fractional_precision
-                         or other_fractional_seconds !=
-                         _scale_to_precision(self.value.microsecond, self_fractional_precision)):
-                    return False
-                if (self_fractional_seconds is not None and other_fractional_seconds is None) and \
-                        (self_fractional_precision != other_fractional_precision
-                         or self_fractional_seconds !=
-                         _scale_to_precision(other.value.microsecond, other_fractional_precision)):
+
+            if isinstance(self.value, Timestamp) and isinstance(other.value, Timestamp):
+                self_fractional_seconds = getattr(self.value, TIMESTAMP_FRACTIONAL_SECONDS_FIELD, None)
+                other_fractional_seconds = getattr(other.value, TIMESTAMP_FRACTIONAL_SECONDS_FIELD, None)
+                if self_fractional_seconds != other_fractional_seconds:
                     return False
 
             if isinstance(self.value, datetime):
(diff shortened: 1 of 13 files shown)


amazon/ion/writer_text.py
```diff
@@ -174,9 +174,20 @@ def _bytes_datetime(dt):
 
     fractional_precision = getattr(original_dt, TIMESTAMP_FRACTION_PRECISION_FIELD, MICROSECOND_PRECISION)
     if fractional_precision:
-        fractional = dt.strftime('%f')
-        assert len(fractional) == MICROSECOND_PRECISION
-        if fractional[fractional_precision:] != ('0' * (MICROSECOND_PRECISION - fractional_precision)):
+        if fractional_precision <= MICROSECOND_PRECISION:
```
tgregg: `fractional_precision` should always be used when present. If not present, the existing logic can be used.

tests/test_reader_binary.py
```diff
@@ -96,8 +96,6 @@ def _prepend_ivm(params):
 )
 
 _BAD_VALUES = (
-    # Only up to microsecond precision is supported (6 digits).
-    (b'\x69\xC0\x81\x81\x81\x80\x80\x80\xC7\x01', 'OVERFLOWING TIMESTAMP PRECISION'),
```
tgregg: Instead of just removing this, let's move it to `_TOP_LEVEL_VALUES` and verify that it is written with full precision.

amazon/ion/core.py
```diff
@@ -144,6 +144,7 @@ def __eq__(self, other):
                 # Special case for timestamps to capture equivalence over precision.
                 self_precision = getattr(self.value, TIMESTAMP_PRECISION_FIELD, None)
                 other_precision = getattr(other.value, TIMESTAMP_PRECISION_FIELD, None)
+                self_fractional_seconds = getattr(other.value, FRACTIONAL_SECONDS, None)
```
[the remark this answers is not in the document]
avneet-toor-bq: Addressed, but please review the changes to ensure I interpreted this correctly.

amazon/ion/core.py
```diff
@@ -366,6 +367,7 @@ def includes_second(self):
 
 TIMESTAMP_PRECISION_FIELD = 'precision'
 TIMESTAMP_FRACTION_PRECISION_FIELD = 'fractional_precision'
+FRACTIONAL_SECONDS = 'fractional_seconds'
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -375,27 +377,28 @@ class Timestamp(datetime):
     Notes:
         The ``precision`` field is passed as a keyword argument of the same name.
```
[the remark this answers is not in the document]
avneet-toor-bq: Done, please review.

amazon/ion/core.py
```diff
@@ -375,27 +377,28 @@ class Timestamp(datetime):
...
         precision = None
         fractional_precision = None
+        fractional_seconds = None
         if TIMESTAMP_PRECISION_FIELD in kwargs:
             precision = kwargs.get(TIMESTAMP_PRECISION_FIELD)
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_PRECISION_FIELD]
         if TIMESTAMP_FRACTION_PRECISION_FIELD in kwargs:
             fractional_precision = kwargs.get(TIMESTAMP_FRACTION_PRECISION_FIELD)
-            if fractional_precision is not None and 1 > fractional_precision > MICROSECOND_PRECISION:
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -469,10 +473,17 @@ def timestamp(year, month=1, day=1,
         if fractional_precision is not None:
             raise ValueError('Cannot have fractional precision without a fractional component.')
 
+    if fractional_precision is not None and fractional_precision > MICROSECOND_PRECISION:
```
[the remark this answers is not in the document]
avneet-toor-bq: Removed. No rounding or truncation done here.

amazon/ion/reader_binary.py
```diff
@@ -714,15 +714,23 @@ def parse_timestamp():
             else:
                 fractional_precision = -1 * fraction_exponent
                 if fractional_precision > MICROSECOND_PRECISION:
-                    raise IonException(
-                        'Timestamp has a fractional component %s with greater than microsecond precision.' % fraction)
-                microsecond = fraction.scaleb(MICROSECOND_PRECISION).to_integral_value()
+                    microsecond = fraction.scaleb(fractional_precision).to_integral_value()
```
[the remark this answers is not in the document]
avneet-toor-bq: Addressed, but please review the changes I've done to fraction.

amazon/ion/reader_text.py
```diff
@@ -928,20 +928,22 @@ def parse():
...
                 fraction_digits = len(fraction)
                 if fraction_digits > MICROSECOND_PRECISION:
+                    has_unnecessary_padding = True
                     for digit in fraction[MICROSECOND_PRECISION:]:
                         if digit != _ZERO:
-                            raise ValueError('Only six significant digits supported in timestamp fractional. Found %s.'
-                                             % (fraction,))
-                    fraction_digits = MICROSECOND_PRECISION
-                    fraction = fraction[0:MICROSECOND_PRECISION]
+                            has_unnecessary_padding = False
```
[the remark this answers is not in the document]
avneet-toor-bq: Done. Microsecond and fractional_precision are now None, while fractional_seconds=microsecond

amazon/ion/writer_binary_raw.py
```diff
@@ -263,19 +253,22 @@ def _serialize_timestamp(ion_event):
         length += _write_varuint(value_buf, dt.minute)
     if precision.includes_second:
         length += _write_varuint(value_buf, dt.second)
-        coefficient = dt.microsecond
+        coefficient = getattr(ion_event.value, FRACTIONAL_SECONDS, dt.microsecond)
+        if coefficient is None:
+            coefficient = dt.microsecond
         fractional_precision = getattr(ion_event.value, TIMESTAMP_FRACTION_PRECISION_FIELD, MICROSECOND_PRECISION)
```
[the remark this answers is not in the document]
avneet-toor-bq: I've addressed this, but check if the change done is the one you intended it to be.

amazon/ion/writer_text.py
avneet-toor-bq: Could you clarify this? I believe the only way this existing logic can be used is if fractional_precision is present, as seen in line 176?

tgregg: Sorry, I meant `fractional_seconds` should always be used when present. And since we're enforcing at Timestamp construction that it will be present whenever there are fractional seconds, the existing logic can be replaced by logic that writes the `fractional_seconds` in the proper form.

amazon/ion/core.py
```diff
@@ -375,27 +392,52 @@ class Timestamp(datetime):
...
-                                 % (fractional_precision, MICROSECOND_PRECISION,))
+                                 'which is less than 1.' % fractional_precision)
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTION_PRECISION_FIELD]
+        if TIMESTAMP_FRACTIONAL_SECONDS_FIELD in kwargs:
+            fractional_seconds = kwargs.get(TIMESTAMP_FRACTIONAL_SECONDS_FIELD)
+            # Make sure we mask this before we construct the datetime.
+            del kwargs[TIMESTAMP_FRACTIONAL_SECONDS_FIELD]
+
+        if args[6] is not None and fractional_precision is not None and fractional_seconds is not None:
```
tgregg: Please name `6` something like `MICROSECOND_ARGUMENT_INDEX`. Might as well also assign the value of `args[MICROSECOND_ARGUMENT_INDEX]` to a variable since it's used several times unconditionally.

Can the value of `microsecond` also occur only in `kwargs`, or is it always present at index 6 in `args`? What if all of the default arguments are used, or if microseconds is passed in using a named parameter?

Also, it looks like we'll need to set `fractional_precision` before this if `microsecond` is set but `fractional_precision` isn't. Or should we raise an error if one is present and the other isn't?

amazon/ion/core.py
```diff
@@ -375,27 +392,52 @@ class Timestamp(datetime):
...
+            if args[6] != fractional_seconds:
```
tgregg: There needs to be some conversion here since `args[6]` is an integer number of microseconds and `fractional_seconds` is a `Decimal` in the range [0, 1).

tgregg: There needs to be some conversion here since `args[6]` is an integer number of microseconds and `fractional_seconds` is a `Decimal` in the range [0, 1).

amazon/ion/core.py
```diff
@@ -375,27 +392,52 @@ class Timestamp(datetime):
...
+                raise ValueError('Microseconds and fractional seconds are not equivalent.')
+
+        if args[6] is not None and fractional_precision is not None and fractional_seconds is None:
+            fractional_seconds = args[6]
```
tgregg: This needs to be a Decimal created from `args[6]` scaled to `-fractional_precision`.

amazon/ion/core.py
```diff
@@ -375,27 +392,52 @@ class Timestamp(datetime):
...
+            fractional_seconds = int(str(fractional_seconds)[0:fractional_precision])
+
+        if args[6] is None and fractional_seconds is not None:
+            fractional_seconds = int(fractional_seconds)
```
tgregg: `fractional_seconds` should remain unchanged. We need to set `microsecond` to the integral value of the `fractional_seconds` `Decimal` scaled to the minimum of its precision and `MICROSECOND_PRECISION`. I'd use `Decimal` methods for this.

amazon/ion/core.py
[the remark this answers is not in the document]
tgregg: Please update this docstring to explain the behavior. Additional comments on the implementation are inline.

amazon/ion/writer_binary_raw.py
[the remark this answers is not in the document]
tgregg: Resolving this thread in favor of my comments on the next revision.

amazon/ion/reader_text.py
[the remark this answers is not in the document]
tgregg: Resolving this thread in favor of my comments on the next revision.

amazon/ion/reader_binary.py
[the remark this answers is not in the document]
tgregg: Resolving this thread in favor of my comments on the next revision.

amazon/ion/core.py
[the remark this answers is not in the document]
tgregg: Resolving this thread in favor of my comments on the next revision.

amazon/ion/core.py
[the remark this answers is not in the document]
avneet-toor-bq: Updated docstrings.

amazon/ion/writer_text.py
avneet-toor-bq: Done. Please review.

amazon/ion/core.py
```diff
@@ -144,11 +144,27 @@ def __eq__(self, other):
...
                 if self_precision != other_precision \
                         and not ((self_precision is None and other_precision is TimestampPrecision.SECOND) or
-                                     (self_precision is TimestampPrecision.SECOND and other_precision is None)):
+                                 (self_precision is TimestampPrecision.SECOND and other_precision is None)):
                     # The absence of precision indicates a naive datetime, which always has SECOND precision.
                     return False
+                if self_fractional_seconds is not None and other_fractional_seconds is not None \
+                        and self_fractional_seconds != other_fractional_seconds:
+                    return False
+                if (self_precision is None and other_fractional_seconds is not None) and \
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -144,11 +144,27 @@ def __eq__(self, other):
...
+                        (other_fractional_precision != MICROSECOND_PRECISION
+                         or other_fractional_seconds != datetime(self.value).microsecond):
+                    return False
+                if (self_precision is not None and other_fractional_seconds is None) and \
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -144,11 +144,27 @@ def __eq__(self, other):
...
-                                     (self_precision is TimestampPrecision.SECOND and other_precision is None)):
+                                 (self_precision is TimestampPrecision.SECOND and other_precision is None)):
                     # The absence of precision indicates a naive datetime, which always has SECOND precision.
                     return False
+                if self_fractional_seconds is not None and other_fractional_seconds is not None \
+                        and self_fractional_seconds != other_fractional_seconds:
+                    return False
+                if (self_precision is None and other_fractional_seconds is not None) and \
+                        (other_fractional_precision != MICROSECOND_PRECISION
+                         or other_fractional_seconds != datetime(self.value).microsecond):
```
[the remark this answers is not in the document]
avneet-toor-bq: Done, please review.

amazon/ion/core.py
```diff
@@ -468,11 +510,10 @@ def timestamp(year, month=1, day=1,
         microsecond = 0
         if fractional_precision is not None:
             raise ValueError('Cannot have fractional precision without a fractional component.')
-
     return Timestamp(
         year, month, day,
         hour, minute, second, microsecond,
-        tz, precision=precision, fractional_precision=fractional_precision
+        tz, precision=precision, fractional_precision=fractional_precision, fractional_seconds=microsecond
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/reader_binary.py
```diff
@@ -714,15 +715,15 @@ def parse_timestamp():
             else:
                 fractional_precision = -1 * fraction_exponent
                 if fractional_precision > MICROSECOND_PRECISION:
-                    raise IonException(
-                        'Timestamp has a fractional component %s with greater than microsecond precision.' % fraction)
-                microsecond = fraction.scaleb(MICROSECOND_PRECISION).to_integral_value()
+                    fraction = fraction.scaleb(fractional_precision).to_integral_value()
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/reader_text.py
```diff
@@ -927,21 +927,12 @@ def parse():
...
-                else:
-                    fraction.extend(_ZEROS[MICROSECOND_PRECISION - fraction_digits])
                 microsecond = int(fraction)
         return timestamp(
             year, month, day,
-            hour, minute, second, microsecond,
+            hour, minute, second, None,
             off_hour, off_minutes,
-            precision=precision, fractional_precision=fraction_digits
+            precision=precision, fractional_precision=None, fractional_seconds=microsecond
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/simple_types.py
```diff
@@ -175,6 +175,8 @@ def _to_constructor_args(ts):
         except AttributeError:
             fractional_precision = MICROSECOND_PRECISION
         kwargs[TIMESTAMP_FRACTION_PRECISION_FIELD] = fractional_precision
+        fractional_seconds = getattr(ts, TIMESTAMP_FRACTIONAL_SECONDS_FIELD, ts.microsecond)
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/writer_binary_raw.py
```diff
@@ -263,19 +253,23 @@ def _serialize_timestamp(ion_event):
         length += _write_varuint(value_buf, dt.minute)
     if precision.includes_second:
         length += _write_varuint(value_buf, dt.second)
-        coefficient = dt.microsecond
-        fractional_precision = getattr(ion_event.value, TIMESTAMP_FRACTION_PRECISION_FIELD, MICROSECOND_PRECISION)
+        coefficient = getattr(ion_event.value, TIMESTAMP_FRACTIONAL_SECONDS_FIELD, dt.microsecond)
```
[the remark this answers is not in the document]
avneet-toor-bq: Done, please review.

amazon/ion/core.py
avneet-toor-bq: Done.

avneet-toor-bq: Done.

amazon/ion/core.py
avneet-toor-bq: Done.

amazon/ion/core.py
avneet-toor-bq: Done, please review.

tests/test_reader_binary.py
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -374,28 +406,85 @@ class Timestamp(datetime):
...
     """
-    __slots__ = [TIMESTAMP_PRECISION_FIELD, TIMESTAMP_FRACTION_PRECISION_FIELD]
+    __slots__ = [TIMESTAMP_PRECISION_FIELD, TIMESTAMP_FRACTION_PRECISION_FIELD, TIMESTAMP_FRACTIONAL_SECONDS_FIELD]
 
     def __new__(cls, *args, **kwargs):
         precision = None
         fractional_precision = None
+        fractional_seconds = None
+        MICROSECOND_ARGUMENT_INDEX = 6
+        DATETIME_MICROSECONDS = None
```
almann: Style nit: these should be lower snake case not upper case.

amazon/ion/core.py
```diff
@@ -374,28 +406,85 @@ class Timestamp(datetime):
...
+                        MICROSECOND_PRECISION,
+                        rounding="ROUND_DOWN")
+                    args = tuple(lst)
+                else:
+                    fractional_seconds = scale_to_precision(DATETIME_MICROSECONDS, MICROSECOND_PRECISION)
+
+            if DATETIME_MICROSECONDS is None and fractional_seconds is not None:
+                lst = list(args)
+                integral_fractional_seconds = fractional_seconds * (10 ** MICROSECOND_PRECISION)
+                lst[MICROSECOND_ARGUMENT_INDEX] = int(integral_fractional_seconds)
```
almann: There is some "don't repeat yourself" going on here, factor out this rounding into its own function, the wall of conditionals here is really hard to follow and that would help it.

tests/test_reader_binary.py
```diff
@@ -187,6 +185,11 @@ def _bad_params():
         b'\x69\x43\xA4\x0F\xE0\x82\x82\x87\x80\x9E',
         e_timestamp(_ts(2016, 2, 2, 0, 0, 30, off_hours=-7, precision=_PREC_SECOND))
     ),
+    (
+        b'\x69\xC0\x81\x81\x81\x80\x80\x80\xC7\x01',
+        e_timestamp(_ts(1, 1, 1, 0, 0, 0, None, precision=_PREC_SECOND, fractional_precision=7,
+                        fractional_seconds=Decimal('1e-7')))
```
almann: This is what I mean about `fractional_precision` and `fractional_seconds`, it seems you don't need both--as the decimal encodes the precision even for zero.  E.g.

```
>>> d = Decimal('1e-7')
>>> d.as_tuple().exponent
-7
```

Am I missing something?

tests/test_reader_text.py
```diff
@@ -789,6 +788,30 @@ def _good_container(start, end, *events):
...
+    (
+        b'2000-01-01T00:00:00.1234567Z',
+        e_timestamp(_ts(
+            2000, 1, 1, 0, 0, 0, 1234567, off_hours=0, off_minutes=0, precision=_tp.SECOND, fractional_precision=7
+        ))
+    ),
+    (
+        b'2000-01-01T00:00:00.1234567800Z',
+        e_timestamp(_ts(
+            2000, 1, 1, 0, 0, 0, 1234567800, off_hours=0, off_minutes=0, precision=_tp.SECOND, fractional_precision=10
```
almann: Shouldn't this case (and cases like it) have a `fractional_seconds` field?

tests/test_simple_types.py
```diff
@@ -49,7 +49,7 @@ def __str__(self):
     (IonPyInt, e_int(2 ** 64 + 1)),
     (IonPyFloat, e_float(1e0)),
     (IonPyDecimal, e_decimal(Decimal('1.1'))),
-    (IonPyTimestamp, e_timestamp(datetime(2012, 1, 1))),
+    (IonPyTimestamp, e_timestamp(Timestamp(2012, 1, 1))),
```
almann: Why did the above need to change?

tests/test_timestamp.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
almann: Minor: should be 2019 :)

tests/test_timestamp.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class _P(record('timestamps', 'expected_microseconds')):
+    def __str__(self):
+        return self.desc
+
+
+MISSING_MICROSECOND = _P(
+    timestamps=[Timestamp(
+            2011, 1, 1,
```
almann: Minor: the indentation in this file is all over the place...

tests/test_timestamp.py
```diff
@@ -0,0 +1,138 @@
...
+            ),
+            Timestamp(
+            2011, 1, 1,
+            0, 0, 0, None,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=Decimal('0.123456')
+            ),
+            Timestamp(
+            2011, 1, 1,
+            0, 0, 0, None,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=Decimal('0.123456789')
```
almann: What is the implication of `fractional_seconds` having more precision that `fractional_precision`?  This distinction needs to be documented as this surprised me...

tests/test_timestamp.py
```diff
@@ -0,0 +1,138 @@
...
+        assert item.timestamps[i].fractional_seconds == item.expected_fractional_precision[i]
+
+
+def test_nonequivalent_microsecond_and_fractional_seconds():
+    with pytest.raises(ValueError):
+        Timestamp(
+            2011, 1, 1,
+            0, 0, 0, 123456,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=0
+        )
```
almann: Minor: I would collapse these into one set of data driven tests, but it isn't a big deal to have them each spelled out.

tests/test_timestamp.py
```diff
@@ -0,0 +1,138 @@
...
+            2011, 1, 1,
+            0, 0, 0, None,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=Decimal('0.000123')
+            ),
+            Timestamp(
+            2011, 1, 1,
+            0, 0, 0, None,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=Decimal('0.123000')
+            )],
+    expected_microseconds=[0, 123456, 123456, 123, 123000]
```
almann: These expectations should be part of each "case" that you have, not a separate list.

tests/test_timestamp.py
```diff
@@ -0,0 +1,138 @@
...
+            2011, 1, 1,
+            0, 0, 0, 123,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=None
+            ),
+            Timestamp(
+            2011, 1, 1,
+            0, 0, 0, 123000,
+            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=None
+            )],
+    expected_fractional_precision=[0, Decimal('0.123456'), Decimal('0.000123'), Decimal('0.123000')]
```
almann: Ditto regarding the record.

I think you are misunderstanding what we're doing with the *parameter* data structures, they are to represent an encapsulation of a test case as data.  So instead of having one *parameter* with two lists, you should have a tuple of *parameters* that you feed in to the test case where *each* parameter is the description of the test case.  In your case the `Timestamp` you constructed and some fields you're predicating on.

tests/test_vectors.py
```diff
@@ -96,7 +96,6 @@ def _open(file):
     # TEXT:
     _good_file(u'subfieldVarUInt.ion'),  # TODO amzn/ion-python#34
     _good_file(u'subfieldVarUInt32bit.ion'),  # TODO amzn/ion-python#34
-    _equivs_file(u'timestampsLargeFractionalPrecision.ion'),  # TODO amzn/ion-python#35
```
almann: Why did this get removed?

amazon/ion/core.py
```diff
@@ -103,6 +104,18 @@ def is_stream_signal(self):
         return self < 0
 
 
+def scale_to_precision(fractional_seconds, fractional_precision):
```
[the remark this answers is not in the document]
avneet-toor-bq: It's not. I changed `scale_to_precision` to use `Decimal.scaleb()`instead of the previous math formula I was using. I think using the `scale_to_precision` method to perform the scaleb action is beneficial, as it makes the code look cleaner.

amazon/ion/core.py
```diff
@@ -430,10 +516,11 @@ def adjust_from_utc_fields(*args, **kwargs):
             adjusted.hour,
             adjusted.minute,
             adjusted.second,
-            adjusted.microsecond,
+            None,
```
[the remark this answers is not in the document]
avneet-toor-bq: No rounding or truncation occurs outside the Timestamp constructer, this was the vision set in this comment: https://github.com/amzn/ion-python/pull/100#discussion_r325348662
Within the Timestamp constructor, if `microseconds` is None while `fractional_seconds` is defined, then we set the value of `microseconds` to the truncated `fractional_seconds`.
Tests to make sure the datetime has at least microsecond precision is not needed I don't think, as microsecond can have less that 6 precision and it always comes back from datetime with microsecond precision?

amazon/ion/core.py
avneet-toor-bq: Done.

amazon/ion/core.py
avneet-toor-bq: Done, factored out into its own function.

tests/test_reader_text.py
avneet-toor-bq: Yes, done.

tests/test_simple_types.py
avneet-toor-bq: I think how it was before was outdated. It expected a datetime back before, but now we get a Timestamp returned to us every time. And this Timestamp has the functionality of both a Timestamp and a datetime.

tests/test_timestamp.py
avneet-toor-bq: Done.

tests/test_timestamp.py
avneet-toor-bq: Fixed. Please review.

tests/test_timestamp.py
avneet-toor-bq: Currently, it's not checked whether or not the `fractional_precision` matches up to `fractional_seconds`. Should this kind of thing be done?

tests/test_timestamp.py
avneet-toor-bq: Done, collapsed into one.

tests/test_timestamp.py
avneet-toor-bq: Done.

tests/test_timestamp.py
avneet-toor-bq: Done.

tests/test_vectors.py
avneet-toor-bq: Removal was outlined in the original issue: https://github.com/amzn/ion-python/issues/35
It's so we can tests large fractional precisions now.

amazon/ion/core.py
```diff
@@ -367,7 +397,7 @@ def includes_second(self):
 
 TIMESTAMP_PRECISION_FIELD = 'precision'
 TIMESTAMP_FRACTION_PRECISION_FIELD = 'fractional_precision'
-FRACTIONAL_SECONDS = 'fractional_seconds'
+TIMESTAMP_FRACTIONAL_SECONDS_FIELD = 'fractional_seconds'
```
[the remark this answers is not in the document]
avneet-toor-bq: By my understanding `fractional_precision` is the unit of measurement for how large `fractional_seconds` is. Good point about the decimal part. Where would this kind of documentation go?

tests/test_simple_types.py
tgregg: Keeping it as `datetime` here helps test that `IonEvent`s and `IonPyTimestamp`s work with both `Timestamp` and `datetime`. I would keep it as a `datetime` unless that causes the test to fail, in which case we should understand why.

Both `IonEvent`s and `IonPyTimestamp`s may be created by the user and passed either to a writer coroutine or `simpleion.dump(s)`, respectively. The user may have a `datetime` object, and that should be accepted by ion-python in all the same places as `Timestamp`.

amazon/ion/core.py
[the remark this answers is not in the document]
almann: It should be in the API documentation for the constructor.  More importantly, you are adding a significantly complex set of parameters, so documenting the invariants around this is paramount.

You have not answered *why* `fractional_seconds` is sufficient to work here by itself--the complexity goes down significantly if you don't need both, but if you do, you should be very explicit as to why.

amazon/ion/core.py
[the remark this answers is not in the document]
almann: That pre-condition to to the argument is not defined anywhere it should be in the docstring (that is API contract).

tests/test_timestamp.py
almann: It depends on why you need both--if you do, then there better be some invariant around both of these fields that is preserved (and tested).

tests/test_vectors.py
almann: Oh is this just the skip list?  Okay, that makes sense then.

amazon/ion/core.py
```diff
@@ -113,7 +113,19 @@ def scale_to_precision(fractional_seconds, fractional_precision):
     Returns:
         Decimal: The scaled value.
     """
-    return Decimal(fractional_seconds) * Decimal(Decimal(10) ** Decimal(-fractional_precision))
+    return Decimal(fractional_seconds).scaleb(-fractional_precision)
+
+
+def round_to_sixth_precision(fractional_seconds):
```
almann: *Minor:* Prefix this with `_` to imply that it is not part of the public surface, this goes for the `scale_to_precision` method above as well (I missed that on first pass).

Also, `round_to_sixth_precision` is not a helpful name, perhaps `_fractional_seconds_to_microseconds`.

tests/test_timestamp.py
almann: No, perhaps I was not clear, I didn't mean collapse your assertions into one test. I meant that you should have a case per each driven as a **parameterized** test.  The problem with your refactoring is that if all of these fail it only looks like one failure.  If you don't want to make it parameterized, keep each test separate.

tests/test_timestamp.py
```diff
@@ -18,121 +18,147 @@
...
-            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=Decimal('0.123000')
-            )],
-    expected_microseconds=[0, 123456, 123456, 123, 123000]
-)
+MISSING_MICROSECOND = [
+    (Timestamp(
+        2011, 1, 1,
+        0, 0, 0, None,
+        precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=0
+        ), 0),
```
almann: So you are using a raw tuple here, but you have the `_P` record class as a named tuple above that you don't appear to be using.  The raw tuples are really hard to read, as a maintainer, what am I supposed to make of index `1` for example?

Why not make a `_P` class that has the `timestamp` and whatever field(s) your testing.  You could make that a dictionary to drive how what the tests are looking for as a way to be more data driven.

amazon/ion/core.py
```diff
@@ -113,7 +113,19 @@ def scale_to_precision(fractional_seconds, fractional_precision):
     Returns:
         Decimal: The scaled value.
     """
-    return Decimal(fractional_seconds) * Decimal(Decimal(10) ** Decimal(-fractional_precision))
+    return Decimal(fractional_seconds).scaleb(-fractional_precision)
```
almann: If the parameter is a `Decimal`, why are you constructing another decimal before invoking `scaleb`?

amazon/ion/core.py
```diff
@@ -113,7 +113,19 @@ def scale_to_precision(fractional_seconds, fractional_precision):
...
+
+
+def round_to_sixth_precision(fractional_seconds):
+    """Scales the ```fractional_seconds``` attribute to a six digit integer.
```
almann: You use the term *round* above and *scale* here, what is really going on?  Its more of a scaling (+ truncation) than it is a rounding, I would make the name and documentation align.

tests/test_timestamp.py
almann: I have follow up below, but resolving this instance.

amazon/ion/core.py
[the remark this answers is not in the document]
almann: I am a bit ambivalent about this, so I think that is fine.

tests/test_timestamp.py
avneet-toor-bq: What did you mean by this? I am using the `_P` class, which has `timestamps` and `expected_value` fields. I am populating these fields using the `event_type_parameters` method.

almann: Reading more carefully, I see what you're doing (and I realize that there are other code paths in the tests doing something similar).

tests/test_simple_types.py
avneet-toor-bq: Upon investigation, this test expects a datetime back but instead we are getting a `Timestamp` back. We are getting a Timestamp back because in the Timestamp constructor, since `microsecond` is defined and `fractional_seconds` is None, we create a new Decimal from the value of `microsecond`, as detailed here: https://github.com/amzn/ion-python/pull/100#discussion_r325348662
What do you think is the right course of action in this case?

amazon/ion/core.py
avneet-toor-bq: Sorry, parameter should be an `int`. So I cast the `int` to a Decimal and then perform the `scaleb` operation on it. Changed parameter description to be an `int` on my end.

amazon/ion/writer_binary_raw.py
```diff
@@ -263,25 +253,39 @@ def _serialize_timestamp(ion_event):
...
+                exponent = -fractional_precision
+                coefficient_fraction_seconds_decimal = (Decimal(coefficient_fraction_seconds *
+                                                                (10**fractional_precision)))
+                coefficient_fraction_seconds_int = int(Decimal(coefficient_fraction_seconds *
+                                                               (10 ** fractional_precision)))
+                if coefficient_fraction_seconds_decimal > coefficient_fraction_seconds_int:
+                    raise ValueError('Error writing event %s. Found timestamp fractional precision of %d digits, '
+                                     'which is less than needed to serialize %d microseconds.'
+                                     % (ion_event, fractional_precision, coefficient_fraction_seconds_decimal))
+                length += _write_decimal_value(value_buf, exponent, coefficient_fraction_seconds_int)
```
tgregg: I think you can factor writer_binary_raw's `_serialize_decimal` lines [177-186](https://github.com/amzn/ion-python/blob/master/amazon/ion/writer_binary_raw.py#L177) into a method and reuse it here. That way you can just pass it the decimal without performing any math here.

amazon/ion/writer_text.py
```diff
@@ -173,15 +173,33 @@ def _bytes_datetime(dt):
...
         fractional = dt.strftime('%f')
         assert len(fractional) == MICROSECOND_PRECISION
+
         if fractional[fractional_precision:] != ('0' * (MICROSECOND_PRECISION - fractional_precision)):
             raise ValueError('Found timestamp fractional with more than the specified %d digits of precision.'
                              % (fractional_precision,))
         fractional = fractional[:fractional_precision]
         tz_string += '.' + fractional
-
+    else:
```
tgregg: Should this be `elif fractional_seconds is not None:`?

amazon/ion/writer_text.py
```diff
@@ -173,15 +173,33 @@ def _bytes_datetime(dt):
...
+        # Since 0.000... is interpreted as 0, must do this in order to have full precision 0's.
+        if fractional_seconds == 0:
+            tz_string += '.' + ('0' * fractional_precision)
+        elif fractional_seconds is not None:
+            if 'e' in str(fractional_seconds):
+                fractional_string = str(Decimal(fractional_seconds))[1:]
```
tgregg: Why `str(Decimal(fractional_seconds))` and not just `str(fractional_seconds)`?

Where do we avoid writing the 'e'?

Does it make more sense (and be cleaner) to just write the coefficient scaled by -exponent (which must be an integer)?

almann:
Based on discussion, here is my take on what the doc string for `Timestamp` should be:

```python
class Timestamp(datetime):
    """Sub-class of :class:`datetime` that supports a precision field; a ``fractional_precision``
    field that specifies the  precision of the``microseconds`` field in :class:`datetime`;
    and a ``fractional_seconds`` field that is a :class:`Decimal` specifying the fractional
    seconds precisely.

    Notes:
        * The ``precision`` field is passed as a keyword argument of the same name.
        
        * The ``fractional_precision`` field is passed as a keyword argument of the same name.
          This field only relates to to the ``microseconds`` field and can be thought of 
          as the number of decimal digits that are significant.  This is an integer that
          that is in the closed interval ``[0, 6]``.  If ``0``, ``microseconds`` must be
          ``0`` indicating no precision below seconds.  This argument is optional and only valid
          when ``microseconds`` is not ``None``.  If the ``microseconds`` specified has more
          precision than this field indicates, then that is an error.  If this is not specified
          it implies ``6`` (i.e. we assume all microseconds are significant).

        * The ``fractional_seconds`` field is passed as a keyword argument of the same name.
          It must be a :class:`Decimal` in the left-closed, right-opened interval of ``[0, 1)``.
          If specified as an argument, ``microseconds`` must be ``None`` **and** ``fractional_precision``
          must not be specified (but can be ``None``).  In addition, if ``microseconds`` is specified
          this argument must not be specified or (but can be ``None``).
        
        * If ``microseconds``, ``fractional_precision``, or ``fractional_seconds`` is specified
          in the constructor, all fields will be present and normalized in the resulting
          :class:`Timestamp` instance.  If the precision of ``fractional_seconds`` is more than
          is capable of being expressed in ``microseconds``, then the ``microseconds`` field
          is truncated to six digits and ``fractional_precision`` is ``6``.

        Consider some examples:
        
        * `2019-10-01T12:45:01Z` would have the following fields set:
          * ``microseconds == 0``, ``fractional_precision == 0``, ``fractional_seconds == Decimal('0')``
        * `2019-10-01T12:45:01.100Z` would have the following fields set:
          * ``microseconds == 100000``, ``fractional_precision == 3``, ``fractional_seconds == Decimal('0.100')``
        * `2019-10-01T12:45:01.123456789Z` would have the following fields set:
          * ``microseconds == 123456``, ``fractional_precision == 6``, ``fractional_seconds == Decimal('0.123456789')``

    Raises:
        ValueError: If any of the preconditions above are violated.
    """
```

amazon/ion/core.py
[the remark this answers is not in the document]
avneet-toor-bq: It was decided not to remove `fractional_precision`. `Timestamp`'s docstring was updated to this contract: https://github.com/amzn/ion-python/pull/100#issuecomment-537212997
with slight modifications in order to make all base unit tests work.

amazon/ion/core.py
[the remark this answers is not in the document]
avneet-toor-bq: Done. API contract properly documented in the `Timestamp` constructor.

tests/test_reader_binary.py
avneet-toor-bq: Changed. It was decided per the contract https://github.com/amzn/ion-python/pull/100#issuecomment-537212997 that `fractional_seconds` does not need `fractional_precision`.

tests/test_simple_types.py
avneet-toor-bq: Fixed, the issue was found to be in core.py's `IonEvent`'s `__eq__` method. Changed back to the original `datetime`.

tests/test_timestamp.py
avneet-toor-bq: We don't need both anymore, as per the contract: https://github.com/amzn/ion-python/pull/100#issuecomment-537212997

tests/test_timestamp.py
avneet-toor-bq: Done. Each test kept separate.

amazon/ion/core.py
avneet-toor-bq: Done. Changed name to `_fractional_seconds_to_microseconds` and prefixed `scale_to_precision` with `_`.

amazon/ion/core.py
avneet-toor-bq: Done. Changed method name to `_fractional_seconds_to_microseconds`.

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
                 raise ValueError('Cannot construct a Timestamp with fractional precision of %d digits, '
-                                 'which is out of the supported range of [1, %d].'
-                                 % (fractional_precision, MICROSECOND_PRECISION,))
+                                 'which is less than 1.' % fractional_precision)
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTION_PRECISION_FIELD]
+        if TIMESTAMP_FRACTIONAL_SECONDS_FIELD in kwargs:
+            fractional_seconds = kwargs.get(TIMESTAMP_FRACTIONAL_SECONDS_FIELD)
+            # Fractional seconds should be a Decimal in the range [0, 1), or None.
+            if fractional_seconds is not None and fractional_seconds > 1:
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
+
+            if fractional_seconds is not None and fractional_precision is None:
+                raise ValueError('Fractional precision cannot be None while fractional seconds is not None.')
+
+            if datetime_microseconds is not None and fractional_precision is not None \
+                    and fractional_seconds is None:
+                if fractional_precision > 6:
+                    fractional_seconds = scale_to_precision(datetime_microseconds, fractional_precision)
+                    lst = list(args)
+                    lst[microsecond_argument_index] = round_to_sixth_precision(fractional_seconds)
```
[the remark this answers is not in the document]
avneet-toor-bq: Removed this logic in favour of ones set by the contract: https://github.com/amzn/ion-python/pull/100#issuecomment-537212997

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
+                    and fractional_seconds is not None:
+                if scale_to_precision(datetime_microseconds, fractional_precision) != fractional_seconds:
+                    raise ValueError('Microseconds and fractional seconds are not equivalent.')
+
+            if fractional_seconds is not None and fractional_precision is None:
+                raise ValueError('Fractional precision cannot be None while fractional seconds is not None.')
+
+            if datetime_microseconds is not None and fractional_precision is not None \
+                    and fractional_seconds is None:
+                if fractional_precision > 6:
```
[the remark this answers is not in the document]
avneet-toor-bq: Done, logic is now simplified based on the contract: https://github.com/amzn/ion-python/pull/100#issuecomment-537212997
Please review.

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
+                args = tuple(lst)
+                datetime_microseconds = args[microsecond_argument_index]
+
+            # If microsecond is still None, set it to 0.
+            if datetime_microseconds is None:
+                lst = list(args)
+                lst[microsecond_argument_index] = 0
+                args = tuple(lst)
+
+        if datetime_microseconds is not None and datetime_microseconds > 999999:
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
+                    fractional_seconds = scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION)
+
+            if datetime_microseconds is None and fractional_seconds is not None:
+                lst = list(args)
+                lst[microsecond_argument_index] = round_to_sixth_precision(fractional_seconds)
+                args = tuple(lst)
+                datetime_microseconds = args[microsecond_argument_index]
+
+            # If microsecond is still None, set it to 0.
+            if datetime_microseconds is None:
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/writer_text.py
avneet-toor-bq: Done.

amazon/ion/writer_text.py
avneet-toor-bq: This logic with the 'e' was not necessary, and it has been removed. I did not find it necessary to write the coefficient scaled by the exponent, as `fractional_seconds` is always a Decimal? Please review.

amazon/ion/writer_binary_raw.py
```diff
@@ -263,25 +253,39 @@ def _serialize_timestamp(ion_event):
...
             if not (coefficient == 0 and exponent >= 0):
                 length += _write_decimal_value(value_buf, exponent, coefficient)
+        else:
+            if coefficient_fraction_seconds is not None:
+                exponent = -fractional_precision
+                coefficient_fraction_seconds_decimal = (Decimal(coefficient_fraction_seconds *
+                                                                (10**fractional_precision)))
+                coefficient_fraction_seconds_int = int(Decimal(coefficient_fraction_seconds *
+                                                               (10 ** fractional_precision)))
+                if coefficient_fraction_seconds_decimal > coefficient_fraction_seconds_int:
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/writer_binary_raw.py
```diff
@@ -263,25 +253,39 @@ def _serialize_timestamp(ion_event):
...
             else:
                 adjusted_fractional_precision = MICROSECOND_PRECISION
+                if fractional_precision > MICROSECOND_PRECISION:
+                    adjusted_fractional_precision = fractional_precision
                 # This optimizes the size of the fractional encoding when the extra precision is not needed.
                 while adjusted_fractional_precision > fractional_precision and coefficient % 10 == 0:
                     coefficient //= 10
                     adjusted_fractional_precision -= 1
-            if adjusted_fractional_precision > fractional_precision or \
-                    coefficient > _TEN_EXP_MINUS_ONE[fractional_precision]:
```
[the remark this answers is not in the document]
avneet-toor-bq: Yes. Added back.

amazon/ion/reader_binary.py
```diff
@@ -699,30 +699,24 @@ def parse_timestamp():
...
+            fraction = None
+            fraction_precision = None
         else:
             fraction = _parse_decimal(buf)
             if fraction < 0 or fraction >= 1:
                 raise IonException(
                     'Timestamp has a fractional component out of bounds: %s' % fraction)
             fraction_exponent = fraction.as_tuple().exponent
             if fraction == 0 and fraction_exponent > -1:
-                # According to the spec, fractions with coefficients of zero and exponents >= zero are ignored.
```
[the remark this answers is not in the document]
avneet-toor-bq: Done. Added back.

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
+            - If microsecond/fractional_precision is specified but fractional_seconds is not, a new Decimal is
+            created from the value of microsecond and truncated to fractional_precision.
+
+            - If fractional_seconds is specified but microsecond is not, microsecond is set
+            to the value of fractional_seconds scaled up to microseconds magnitude.
+
+            - A ValueError is raised when both microsecond and fractional_seconds are specified, but their values are
+            not equivalent.
+
+            - A ValueError is raised when fractional_seconds is specified, but fractional_precision is not.
```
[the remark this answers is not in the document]
avneet-toor-bq: Updated the contract. I think the case of "`fractional_precision` is specified, but `microseconds` is not" is covered in the `Raises:
            ValueError: If any of the preconditions above are violated.`
line. Please review.

amazon/ion/core.py
```diff
@@ -374,28 +418,86 @@ class Timestamp(datetime):
...
+            # Make sure we mask this before we construct the datetime.
+            del kwargs[TIMESTAMP_FRACTIONAL_SECONDS_FIELD]
+        if len(args) > 6:
+            if datetime_microseconds is not None and fractional_precision is not None \
+                    and fractional_seconds is not None:
+                if scale_to_precision(datetime_microseconds, fractional_precision) != fractional_seconds:
+                    raise ValueError('Microseconds and fractional seconds are not equivalent.')
+
+            if fractional_seconds is not None and fractional_precision is None:
+                raise ValueError('Fractional precision cannot be None while fractional seconds is not None.')
```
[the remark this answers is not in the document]
avneet-toor-bq: Yes. Done.

amazon/ion/core.py
```diff
@@ -103,6 +104,30 @@ def is_stream_signal(self):
...
+
+def round_to_sixth_precision(fractional_seconds):
+    """Scales the ```fractional_seconds``` attribute to a six digit integer.
+
+    Args:
+        fractional_seconds (Decimal): The numerical value to scale.
+    Returns:
+        int: The scaled value.
+    """
+    integral_fractional_seconds = fractional_seconds * (10 ** MICROSECOND_PRECISION)
```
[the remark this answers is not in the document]
avneet-toor-bq: Done.

amazon/ion/writer_binary_raw.py
avneet-toor-bq: Done.

amazon/ion/core.py
```diff
@@ -446,53 +461,46 @@ def __new__(cls, *args, **kwargs):
             del kwargs[TIMESTAMP_PRECISION_FIELD]
         if TIMESTAMP_FRACTION_PRECISION_FIELD in kwargs:
             fractional_precision = kwargs.get(TIMESTAMP_FRACTION_PRECISION_FIELD)
-            if fractional_precision is not None and fractional_precision < 1:
+            if fractional_precision is not None and 1 > fractional_precision > MICROSECOND_PRECISION:
```
almann: When `microseconds == 0`, why wouldn't `fractional_precision == 1` be allowed?  That is to say, When `fractional_precision == None`, then it defaults to effectively `fractional_precision == 6` right?

In other words, what is the precision of `microseconds` when this is not specified (as is the case with a normal `datetime`)?

amazon/ion/core.py
```diff
@@ -411,25 +400,51 @@ def includes_second(self):
...
+        Notes:
+            * The ``precision`` field is passed as a keyword argument of the same name.
+
+            * The ``fractional_precision`` field is passed as a keyword argument of the same name.
+              This field only relates to to the ``microseconds`` field and can be thought of
+              as the number of decimal digits that are significant.  This is an integer that
+              that is in the closed interval ``[0, 6]``.  If ``None``, ``microseconds`` must be
+              ``0`` indicating no precision below seconds.  This argument is optional and only valid
+              when ``microseconds`` is not ``None``.  If the ``microseconds`` specified has more
+              precision than this field indicates, then that is an error.
```
almann: This revision to the API is kind of strange, what does it mean to have it not be specified, is that invalid?  `None` and not specified should be the same (i.e. defaulted to `None`), this is also incongruent to how normal `datetime` instances work.

Put another way a normal `datetime` will not have a `fractional_precision` which is congruent to having it `None`.  Are you saying you distinguish between unspecified and `None`, that seems like a very tricky distinction to have for the API contract...

amazon/ion/writer_binary_raw.py
```diff
@@ -226,6 +233,16 @@ def _serialize_lob_value(event, tid):
...
+_TEN_EXP_MINUS_ONE = [
+    -1,
+    9,
+    99,
+    999,
+    9999,
+    99999,
+    999999,
+]
+
```
almann: What is this constant for? Probably warrants a comment and should probably be a `tuple` instead of `list`.

tests/test_reader_text.py
```diff
@@ -788,27 +788,6 @@ def _good_container(start, end, *events):
...
-            fractional_seconds=Decimal('0.1234567')
-        ))
-    ),
-    (
-        b'2000-01-01T00:00:00.1234567800Z',
-        e_timestamp(_ts(
-            2000, 1, 1, 0, 0, 0, 1234567800, off_hours=0, off_minutes=0, precision=_tp.SECOND, fractional_precision=10,
-            fractional_seconds=Decimal('0.1234567800')
-        ))
-    ),
```
almann: You deleted these cases, but aren't they valid (just with different field values)? Or are these test cases redundant?

amazon/ion/core.py
avneet-toor-bq: The logic exists where when  `fractional_precision == None` then `fractional_precision` defaults to 6, but this exists in the `timestamp` method on line [548](https://github.com/amzn/ion-python/pull/100/files#diff-6328c9946663c2b27d43e85f0fa42c3fR548).
Unfortunately, with the way unit tests are set up, for some tests if data just goes through the `Timestamp` constructor with `fractional_precision` being originally `None`, then the test expects the newly constructed `Timestamp` object to have `fractional_precision` of `None`. But if the data go through the `timestamp` function with `fractional_precision` being originally `None`, then the test expects to see `fractional_precision == 6`.
If I try moving the logic to make `fractional_precision == 6` from `timestamp` to `Timestamp`, unit tests start to fail due to the aforementioned reason.

amazon/ion/writer_binary_raw.py
avneet-toor-bq: This constant was here originally, before I started making changes. It checks if `fractional_precision` has too few digits to support the current `microsecond` precision [here.](https://github.com/amzn/ion-python/pull/100/files/c44ce224c1e66cec7c3dd9fc7f8746e447fb73cf..788adafde6ad54312adc3330e2d1abc1edbfe723#diff-d98fe681a3236eae60289fa56a2a0083R286) I removed it before, but we still need it now in case we have a normal `datetime` with no `fractional_seconds`to write.

amazon/ion/writer_text.py
```diff
@@ -173,15 +173,22 @@ def _bytes_datetime(dt):
...
+            if fractional[fractional_precision:] != ('0' * (MICROSECOND_PRECISION - fractional_precision)):
+                raise ValueError('Found timestamp fractional with more than the specified %d digits of precision.'
+                                 % (fractional_precision,))
+            fractional = fractional[:fractional_precision]
+            tz_string += '.' + fractional
+
+    if fractional_seconds == 0:
+        tz_string += '.' + ('0' * fractional_precision)
+    elif fractional_seconds is not None:
+        tz_string += str(fractional_seconds)[1:]
```
tgregg: Is `str(fractional_seconds)` always guaranteed to give a string representation that doesn't contain 'e'? Even for a decimal value with precision of, say, 500?

amazon/ion/writer_text.py
tgregg: Let's continue this discussion on my latest review.

tests/test_timestamp.py
```diff
@@ -0,0 +1,149 @@
...
+        ), 123),
+    (Timestamp(
+        2011, 1, 1,
+        0, 0, 0, None,
+        precision=TimestampPrecision.SECOND, fractional_seconds=Decimal('0.123000')
+        ), 123000)
+]
+
+
+MISSING_FRACTIONAL_PRECISION = [
```
tgregg: This should be `MISSING_FRACTIONAL_SECONDS`, right?

tests/test_timestamp.py
```diff
@@ -0,0 +1,149 @@
...
+            precision=TimestampPrecision.SECOND, fractional_precision=0, fractional_seconds=0
+        )
+
+
+def test_fractional_seconds_greater_than_1():
+    with pytest.raises(ValueError):
+        Timestamp(
+            2011, 1, 1,
+            0, 0, 0, 0,
+            precision=TimestampPrecision.SECOND, fractional_precision=1, fractional_seconds=2
```
tgregg: Wouldn't this also cause an error because both `fractional_seconds` and `fractional_precision` are specified, and we've made them mutually exclusive?

tests/test_timestamp.py
```diff
@@ -0,0 +1,149 @@
...
+        0, 0, 0, 123000,
+        precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=None
+        ), Decimal('0.123000'))
+]
+
+
+@listify
+def event_type_parameters(list_name):
+    print(list_name)
+    for timestamp, expected_val in list_name:
```
tgregg: For me, the indentation of the tuples in the `MISSING_MICROSECOND` and `MISSING_FRACTIONAL_PRECISION` lists made it really hard to see that they each contained two elements. I stared at this for a little while before realizing that it wasn't just a single call to the Timestamp constructor. I recommend putting the two elements at the same indentation level and putting continuation lines at an increased indentation level. Something like:

```python
    (
        Timestamp(
            2011, 1, 1,
            0, 0, 0, 123000,
            precision=TimestampPrecision.SECOND, fractional_precision=6, fractional_seconds=None
        ),
        Decimal('0.123000')
    )
```

amazon/ion/writer_binary_raw.py
tgregg: The fact that it's not a `tuple` is my fault as the original author. I agree that a `tuple` is more suitable.

amazon/ion/core.py
```diff
@@ -389,13 +464,48 @@ def __new__(cls, *args, **kwargs):
...
+        elif not (datetime_microseconds == 0 and fractional_precision is None):
+            if datetime_microseconds is None:
+                if len(args) > 6:
+                    datetime_microseconds = 0
+                    lst = list(args)
+                    lst[microsecond_argument_index] = datetime_microseconds
+                    args = tuple(lst)
+            else:
+                fractional_seconds = str(_scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION))
+                fractional_seconds = Decimal(fractional_seconds[0:(fractional_precision + 2)])
```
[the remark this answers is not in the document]
avneet-toor-bq: It's to make sure we have `fractional_seconds` at the right precision when we compute it from `datetime_microseconds` and `fractional_precision`. 
Example:
Let's say in the timestamp we want microseconds of 0.001000 with precision 3, so 0.001. This value will be stored in `datetime_microseconds` as 1000 and `fractional_precision` will be 3.
We want to compute the corresponding `fractional_seconds`. In line 502, what we get back from `_scale_to_precision` is 0.001000 and we stringify it.
Then in line 503 we truncate this string to `fractional_precision + 2` (to account for the `0.` in the beginning of the string), cast it to Decimal, then we end up with the correct `fractional_seconds` Decimal value of 0.001.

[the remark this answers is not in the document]
tgregg: I see. It may be more succinct and efficient (avoiding the round trip via `str`) to `_scale_to_precision(datetime_microseconds, MICROSECOND_PRECISION).quantize(precision_limit, decimal.ROUND_FLOOR)`, where `precision_limit` is `Decimal('0.' + '0' * (fractional_precision - 1) + '1')`. We can even have a static lookup table for `precision_limit` since `fractional_precision` is guaranteed to be [1, 6] at this point. (See: [Decimal.quantize](https://docs.python.org/2/library/decimal.html#decimal.Decimal.quantize))

amazon/ion/writer_binary_raw.py
```diff
@@ -279,9 +287,13 @@ def _serialize_timestamp(ion_event):
...
                                  % (ion_event, fractional_precision, dt.microsecond))
-            exponent = -adjusted_fractional_precision
-            if not (coefficient == 0 and exponent >= 0):
-                length += _write_decimal_value(value_buf, exponent, coefficient)
+            if coefficient_fraction_seconds is None:
+                exponent = -adjusted_fractional_precision
+                if not (coefficient == 0 and exponent >= 0):
+                    length += _write_decimal_value(value_buf, exponent, coefficient)
+
+        if coefficient_fraction_seconds is not None:
```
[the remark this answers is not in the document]
avneet-toor-bq: Unfortunately, some unit tests rely on line 287 being thrown, when fractional_precision is less than needed to serialize microseconds. Even if `coefficient_fraction_seconds` is present, unit tests still fail because they expect ValueError to be raised so this check needs to be done. This is why I ordered things this way, to work around this. 
Ideally, as a side task, this check should be moved to the Timestamp constructor and unit tests for writer_binary_raw should be updated accordingly.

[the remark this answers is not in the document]
tgregg: Okay, this is definitely one of the things that we should call out for cleanup after this PR is merged.

> Even if `coefficient_fraction_seconds` is present, unit tests still fail because they expect ValueError to be raised so this check needs to be done.

We will need to update these tests that expect failure such that their inputs are always raw `datetime` objects, as `Timestamp` objects should never use `dt.microsecond`.

> Ideally, as a side task, this check should be moved to the Timestamp constructor and unit tests for writer_binary_raw should be updated accordingly.

I agree, this bounds checking should be added to the constructor (and we should probably make note of this in the docstring once that's done). For example, a `ValueError` should be raised if the user tries to construct a `Timestamp` with `microsecond=999999` and `fractional_precision=3` because data would be lost.

amazon/ion/core.py
```diff
@@ -104,27 +104,27 @@ def is_stream_signal(self):
...
 
     Args:
-        fractional_seconds (Decimal): The numerical value to scale.
+        microseconds (int): The numerical value to scale.
         fractional_precision (int): The negative of this numerical value to scale ```fractional_seconds``` by.
     Returns:
         Decimal: The scaled value.
     """
-    return Decimal(fractional_seconds).scaleb(-fractional_precision)
+    return Decimal(microseconds).scaleb(-fractional_precision)
```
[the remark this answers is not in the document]
avneet-toor-bq: - No, I don't think we should drop `fractional_precision` in favour of number of digits in first parameter. 
Think of the case where in the timestamp we want microseconds of 0.001000. This value will be represented in the int parameter we're passing as as 1000.
If we rely on number of digits, the `Decimal` that will get returned will be `0.1000`, which is incorrect.
- Changed first parameter's name from `microseconds` to `value`.
- Also added logic here to check if `value`'s number of significant digits matches that of `fractional_seconds.` I did this change as I knew previously `reader_text.py` was this kind of check and since it is using this `_scale_to_precision()` method too, I thought it'd be good to put this check here.

tests/test_reader_text.py
avneet-toor-bq: Added back, with the appropriate field values.

amazon/ion/core.py
```diff
@@ -389,13 +464,48 @@ def __new__(cls, *args, **kwargs):
...
                 raise ValueError('Cannot construct a Timestamp with fractional precision of %d digits, '
                                  'which is out of the supported range of [1, %d].'
-                                 % (fractional_precision, MICROSECOND_PRECISION,))
+                                 % (fractional_precision, MICROSECOND_PRECISION))
             # Make sure we mask this before we construct the datetime.
             del kwargs[TIMESTAMP_FRACTION_PRECISION_FIELD]
+        if TIMESTAMP_FRACTIONAL_SECONDS_FIELD in kwargs:
+            fractional_seconds = kwargs.get(TIMESTAMP_FRACTIONAL_SECONDS_FIELD)
+            if fractional_seconds is not None:
+                fractional_seconds = Decimal(fractional_seconds)
```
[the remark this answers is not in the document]
avneet-toor-bq: No. Removed.

amazon/ion/core.py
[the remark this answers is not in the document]
avneet-toor-bq: Done. Thank you.

amazon/ion/writer_binary_raw.py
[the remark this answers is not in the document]
avneet-toor-bq: Unfortunately, I cannot add this bounds checking to the `Timestamp` constructor now. The same tests that previously complained a `ValueError` wasn't being thrown now complain that the `ValueError` weren't thrown within `writer_binary_raw`.
After this PR has been merged, bounds checking should be added to `Timestamp` and unit tests need to be updated.

amazon/ion/writer_text.py
avneet-toor-bq: Good catch. Added a check for this, and some logic to handle this scenario contained within a new method called `scientific_notation_to_decimal_string()`

tests/test_timestamp.py
avneet-toor-bq: Right. Changed.

tests/test_timestamp.py
avneet-toor-bq: Correct. Changed this test so that `fractional_precision` is `None` and `fractional_seconds` has a value.

tests/test_timestamp.py
avneet-toor-bq: Fixed.

amazon/ion/writer_binary_raw.py
avneet-toor-bq: Done, changed to tuple.

amazon/ion/core.py
```diff
@@ -104,16 +104,31 @@ def is_stream_signal(self):
...
+    value_string = str(value)
+    if len(value_string) > fractional_precision:
+        for digit in value_string[fractional_precision:]:
+            if digit != '0':
+                raise ValueError('Found timestamp fractional precision of %d digits, '
+                                 'which is less than needed to serialize %d microseconds.'
+                                 % (fractional_precision, value))
+        value_string = value_string[0:fractional_precision]
+        value = int(value_string)
+    return Decimal(value).scaleb(-fractional_precision)
```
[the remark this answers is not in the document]
avneet-toor-bq: I went this route because this exact sort of logic using `__str__` was being used in `reader_text.py` initially. The only difference, however, was in the method that used this logic in `reader_text.py` we didn't cast from `Decimal` to string to use this logic, as within the method parameters a string was being passed.

amazon/ion/writer_text.py
```diff
@@ -188,7 +209,10 @@ def _bytes_datetime(dt):
     if fractional_seconds == 0:
         tz_string += '.' + ('0' * fractional_precision)
     elif fractional_seconds is not None:
-        tz_string += str(fractional_seconds)[1:]
+        if 'E' in str(fractional_seconds):
```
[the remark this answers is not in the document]
avneet-toor-bq: As far as I recall, `Decimal` can be represented as a full-precision decimal or via scientific notation with this [format](https://python-reference.readthedocs.io/en/latest/docs/float/scientific.html). 
And when it comes to casting as string, the Decimal as-is (whether it's represented as full precision decimal like 0.01 or scientific notation like 1E-9) should be retained when cast to string. 
For example, str(1E-9) == '1E-9'
I have accounted for the case where we see 'e' instead of 'E' for the scientific notation representation in the next commit.

almann:
Squashed and merged here:

https://github.com/amzn/ion-python/commits/timestamp-precision

We will use this branch as a staging point for the next PR.

PR #106, opened 2019-10-10
Makes the Timestamp constructor comply with the contract stated in its docstring; simplifies Timestamp writing logic where possible.

tgregg:
*Issue #, if available:*
Continues #100

*Description of changes:*
* In the `Timestamp` docstring:
    * Explicitly states that `fractional_seconds=Decimal(0)` is equivalent to having no precision past seconds. 
    * Clarifies that under all circumstances, a constructed `Timestamp` will have populated values for `microsecond`, `fractional_precision`, and `fractional_seconds`.
* Makes the `Timestamp` constructor comply with the contract laid out in the docstring.
* Allows `microsecond` to be specified as either a positional or a keyword argument.
* Fixes the range checking for `fractional_precision` and `fractional_seconds` and the associated tests.
* Removes all defaulting logic from the `timestamp` function. In previous revisions this defaulting led to incongruity between results of `Timestamp` and `timestamp` given the same arguments.
* Removes validation in several places that is redundant now that it is performed as part of the `Timestamp` constructor contract.
* Writes timestamp fractional seconds using `fractional_seconds` whenever the value is a `Timestamp` since this is guaranteed to be populated after construction.
* Adds some additional test coverage and improves the organization of `test_timestamp.py`.

### DIFF
--- a/tests/writer_util.py
+++ b/tests/writer_util.py
@@ -159,6 +159,53 @@ SIMPLE_SCALARS_MAP_TEXT = {
         (_DT(2016, 1, 1, 12, 34, 12, tzinfo=OffsetTZInfo()), b'2016-01-01T12:34:12.000000Z'),
         (_DT(2016, 1, 1, 12, 34, 12, tzinfo=OffsetTZInfo(timedelta(hours=-7))),
             b'2016-01-01T12:34:12.000000-07:00'),
+        (timestamp(year=1, month=1, day=1, precision=TimestampPrecision.DAY), b'0001-01-01T'),
+        (timestamp(year=1, month=1, day=1, off_minutes=-1, precision=TimestampPrecision.SECOND),
+         b'0001-01-01T00:00:00-00:01'),
+        (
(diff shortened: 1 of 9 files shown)


tests/writer_util.py
```diff
@@ -159,6 +159,53 @@ def _convert_clob_pairs(clob_pairs):
         (_DT(2016, 1, 1, 12, 34, 12, tzinfo=OffsetTZInfo()), b'2016-01-01T12:34:12.000000Z'),
         (_DT(2016, 1, 1, 12, 34, 12, tzinfo=OffsetTZInfo(timedelta(hours=-7))),
             b'2016-01-01T12:34:12.000000-07:00'),
+        (timestamp(year=1, month=1, day=1, precision=TimestampPrecision.DAY), b'0001-01-01T'),
```
tgregg: These were copied from `SIMPLE_SCALARS_MAP_BINARY` below and given expected text values. There's no reason to have this coverage for binary but not text.

PR #109, opened 2019-10-17
Adds option to omit the Ion version marker from text output

### LINKED ISSUE #94: Omit Ion identifier for pretty printing text
When pretty-printing elements from an array of Ion values, and the context makes it clear that each value is Ion, allow the $ion_1_0 prefix to be omitted as it is redundant and adds clutter to the output.

pbcornell:
Resolves #94 

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -49,7 +49,7 @@ def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, skipkeys=
          check_circular=True, allow_nan=True, cls=None, indent=None, separators=None, encoding='utf-8', default=None,
          use_decimal=True, namedtuple_as_object=True, tuple_as_array=True, bigint_as_string=False, sort_keys=False,
          item_sort_key=None, for_json=None, ignore_nan=False, int_as_string_bitcount=None, iterable_as_array=False,
-         tuple_as_sexp=False, **kw):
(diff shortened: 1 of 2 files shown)


amazon/ion/simpleion.py
```diff
@@ -138,14 +138,17 @@ def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, skipkeys=
...
+            Ion Version Marker ($ion_1_0) from the output.  Default: False.
         **kw: NOT IMPLEMENTED
 
     """
 
     raw_writer = binary_writer(imports) if binary else text_writer(indent=indent)
     writer = blocking_writer(raw_writer, fp)
     from_type = _FROM_TYPE_TUPLE_AS_SEXP if tuple_as_sexp else _FROM_TYPE
-    writer.send(ION_VERSION_MARKER_EVENT)  # The IVM is emitted automatically in binary; it's optional in text.
+    if not binary and not omit_version_marker:
```
tgregg: Shouldn't this be `if binary or not omit_version_marker:`?

I mean, the tests pass so the current logic clearly works, but it looks weird to me. Are we relying on the fact that under the hood the binary writer always writes the ION_VERSION_MARKER_EVENT without being explicitly told to?

pbcornell: Agreed, thanks for catching this.  Fixing in https://github.com/amzn/ion-python/pull/111

Issue #126, opened 2020-04-09
AttributeError: Can't pickle local object '_ion_type_for.<locals>.IonPyValueType'


yosoufe:
Hello, 
I need to use multiprocessing module of python and I would like to pass `IonPy*` types between processes but it seems those objects are not picklable. 

In order to reproduce, the following code
```py
from amazon.ion import simpleion as ion
import multiprocessing

def process_main(return_queue):
    return_queue.put(ion.IonPyText("test text"))
    print("DO Something")

if __name__== "__main__":
    ctx = multiprocessing.get_context('spawn')
    return_queue = ctx.Queue()
    p = ctx.Process(target = process_main, args=(return_queue,))
    p.start()
    p.join(10)
    p.terminate()
```
is creating the following output message:
```
DO Something
Traceback (most recent call last):
  File "/usr/lib/python3.5/multiprocessing/queues.py", line 241, in _feed
    obj = ForkingPickler.dumps(obj)
  File "/usr/lib/python3.5/multiprocessing/reduction.py", line 50, in dumps
    cls(buf, protocol).dump(obj)
AttributeError: Can't pickle local object '_ion_type_for.<locals>.IonPyValueType'
```

It can be because the `IonPyValueType` class is being defined [inside a function](https://github.com/amzn/ion-python/blob/master/amazon/ion/simple_types.py#L136) rather than in top level of the module and accoding to [python documentation](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled) that would make it unpicklable.

I would be grateful of any feedback? Am I doing something wrong?  Is there any workaround that I am missing?
Thanks

yosoufe:
[This change](https://github.com/yosoufe/ion-python/commit/bfdabeb38fc89d98e10d5df6278f49e450b8ff44) seems to solve the above problem but I have another problem that I want to reproduce it. 

With the above change,
I have IonPy* attribute for a custom objects and during the unpickling I get the following error
```
TypeError: __new__() missing 1 required positional argument: 'value'
```

yosoufe:
Ok I think I resolved the bug above as well in [this commit](https://github.com/yosoufe/ion-python/commit/69f2fbaa346a92f1d289d914aa09f0681aa25d20). `Enum` classed needed `__getnewargs__` function definition. I would run the tests and if it was ok I will open a PR. Let me know if you see something wrong here. Thanks

yosoufe:
Closing. Resolved by the merged PR above.

Issue #61, opened 2018-05-09
Deep Copy fails / Save to DynamoDB fails


pettijohn:
I can't save an Ion payload to DynamoDB with Boto3. It looks like Boto3 makes a deep copy prior to saving, which throws. 

```
import boto3
import copy
from decimal import Decimal
from amazon.ion import simpleion
from io import StringIO


d = { "value": Decimal('1.1') }
payload = simpleion.dumps(d)
payloadValue = simpleion.load(StringIO(payload))
c = copy.deepcopy(payloadValue) # This fails
table = boto3.resource('dynamodb').Table('MyTable')
table.put_item(Item=payloadValue) # This fails because it makes a deep copy
```

Should an Ion struct be able to be deep copied? Stack below.

```
TypeError: __new__() missing 1 required positional argument: 'value'

Traceback (most recent call last):
  File "c:\Projects\\lambda\lambda_function.py", line 32, in <module>
    bug()
  File "c:\Projects\\lambda\lambda_function.py", line 29, in bug
    c = copy.deepcopy(payloadValue)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 274, in _reconstruct
    y = func(*args)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 273, in <genexpr>
    args = (deepcopy(arg, memo) for arg in args)
  File "C:\Users\\AppData\Local\Programs\Python\Python36\lib\copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
[...]

PR #127, opened 2020-04-09
Making simple_ion types picklable

yosoufe:
Using `type` to create simple ion classes to make them picklable and also add the __getnewargs__ to Enum class to make the ion_type enums picklable

*Issue #, if available:* https://github.com/amzn/ion-python/issues/126

The error message I was getting before this change was similar to https://github.com/amzn/ion-python/issues/61 so this MR may also resolve that issue. But I do not know and I have not tested it.

*Description of changes:*
Because the `IonPyValueType` is defined in a function and not in the module scope so it is not picklable 
according to [python documentation](https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled) and it seems `Enum` type was needing `__getnewargs__` function. I really need somebody who knows python well review my implementation of `__getnewargs__`. I am not sure what exactly `__getnewargs__` does and how it should be implemented. But it seems it is working. The code I tried and tested this PR with is:

```py
from amazon.ion import simpleion as ion
import multiprocessing
import pickle
from collections import OrderedDict

def process_1(return_queue):
    ion_objects = [
        (ion.IonPyList([1,2,3]),                            ion.IonType.LIST),
        (ion.IonPyDict(OrderedDict([("Hi","There")])),      ion.IonType.STRUCT),
        (ion.IonPyDict({"Hi":ion.IonPyFloat(1.0)}),         ion.IonType.STRUCT),
        (ion.IonPyDecimal(1),                               ion.IonType.DECIMAL),
        (ion.IonPyText("Test"),                             ion.IonType.STRING),
        (ion.IonPyBool(True),                               ion.IonType.BOOL),
        (ion.IonPySymbol("symbol_test", None),              ion.IonType.SYMBOL)
    ]
    for obj, obj_type in ion_objects:
        ion_object = obj
        ion_object.ion_type = obj_type
        pickle.loads(pickle.dumps(ion_object))  # direct pickling
        return_queue.put(ion_object)            # pickling is happening under the hood

class Custom:
    def __init__(self):
        self.symbolTest = ion.IonPySymbol("SOME Symbol", None)
        self.symbolTest.ion_type = ion.IonType.SYMBOL

def process_2(return_queue):
    c = Custom()
    return_queue.put(c)

if __name__== "__main__":
    for proc in [process_1, process_2]:
        print(proc.__name__)
        ctx = multiprocessing.get_context('spawn')
        return_queue = ctx.Queue()
        p = ctx.Process(target = proc, args=(return_queue,))
        p.start()
        p.join(10)
        p.terminate()
        while not return_queue.empty():
            v = return_queue.get(block=False)
            print(type(v), v)
```
I also ran the unit tests. All passed.

We need to pickle because multiprocessing is pickling objects to pass them to outside of the process. That is the main goal of this PR.

### DIFF
--- a/amazon/ion/simple_types.py
+++ b/amazon/ion/simple_types.py
@@ -133,26 +133,15 @@ class _IonNature(object):
                                   annotations=self.ion_annotations, depth=depth)
 
 
-def _ion_type_for(name, base_cls):
-    class IonPyValueType(base_cls, _IonNature):
-        def __init__(self, *args, **kwargs):
-            super(IonPyValueType, self).__init__(*args, **kwargs)
-
-    IonPyValueType.__name__ = name
-    return IonPyValueType
-
-
 if six.PY2:
-    IonPyInt = _ion_type_for('IonPyInt', long)
+    IonPyInt = type('IonPyInt', (long,_IonNature), {'__module__':__name__})
 else:
-    IonPyInt = _ion_type_for('IonPyInt', int)
-
-
+    IonPyInt = type('IonPyInt', (int,_IonNature), {'__module__':__name__})
 IonPyBool = IonPyInt
-IonPyFloat = _ion_type_for('IonPyFloat', float)
-IonPyDecimal = _ion_type_for('IonPyDecimal', Decimal)
-IonPyText = _ion_type_for('IonPyText', six.text_type)
-IonPyBytes = _ion_type_for('IonPyBytes', six.binary_type)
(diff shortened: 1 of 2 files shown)


yosoufe:
I found that with less change we can make the types picklable. This is the commit that works for me with less change: 
https://github.com/yosoufe/ion-python/commit/2097fdc2436fc4436eef8c2cf041a7925daa0f20

I will close this PR and open another one for the other change.

yosoufe:
I opened this PR instead: https://github.com/amzn/ion-python/pull/128/files

Issue #132, opened 2020-06-08
Decimal example where value is being changed when using `loads`


oscar26:
Hi,

I'm writing a little utility for converting Ion objects to Python objects and while testing it, I've found a particular example of a `Decimal` object not being converted properly:

```python
from decimal import Decimal
from amazon.ion.simpleion import dumps, loads

decimal_val = Decimal('1005826281919371473355538432.5')
ion_value = loads(dumps(decimal_val))
print(ion_value)
```

`1005826281919371473355538432` is printed. Note the missing decimal part. `dumps` apparently doesn't modify the value as in here:

```python
print(dumps(decimal_val, binary=False, omit_version_marker=True))
```
The value is printed correctly: `1005826281919371473355538432.5`, so the problem may be somewhere inside `loads`.

I've found this example by using the [Hypothesis](https://hypothesis.readthedocs.io/en/latest/index.html) testing library and there are several other cases where my tests pass just fine.

Additional info:
* Python version: 3.7.7
* `amazon.ion` version: 0.5.0 (I'm using this version as `pyqldb` has not bumped its `amazon.ion` dependency to 0.6.0 but the problem is present in this version as well)

Thank you for your time and effort!

oscar26:
Hi Peter,

Indeed, interesting it is. Decided to try your snippet on Python `3.5.9`, `3.6.10`, `3.8.3` and `3.9.0b2` with no success.

Also, it seems like `scaleb()` is performing a rounding somewhere:

```python
>>> Decimal('1005826281919371473355538432.0').scaleb(0)
Decimal('1005826281919371473355538432')
>>> Decimal('1005826281919371473355538432.1').scaleb(0)
Decimal('1005826281919371473355538432')
>>> Decimal('1005826281919371473355538432.2').scaleb(0)
Decimal('1005826281919371473355538432')
>>> Decimal('1005826281919371473355538432.3').scaleb(0)
Decimal('1005826281919371473355538432')
>>> Decimal('1005826281919371473355538432.4').scaleb(0)
Decimal('1005826281919371473355538432')
>>> Decimal('1005826281919371473355538432.5').scaleb(0)
Decimal('1005826281919371473355538432')
>>> Decimal('1005826281919371473355538432.6').scaleb(0)
Decimal('1005826281919371473355538433')
>>> Decimal('1005826281919371473355538432.7').scaleb(0)
Decimal('1005826281919371473355538433')
>>> Decimal('1005826281919371473355538432.8').scaleb(0)
[...]

PR #133, opened 2020-06-12
Fix: use localcontext to handle decimals of any precision

oscar26:
*Issue #132*

*Description of changes:*
Use [`localcontext`](https://docs.python.org/3.7/library/decimal.html#decimal.localcontext) from to adjust the precision and allow any number of type `decimal` to be correctly de/serialized from/to Ion.

Also, I don't know where and how to add tests to prevent potential regression bugs. Locally I ran `python setup.py test` and all tests passed and in addition I used the following snippet:

```python
from decimal import Decimal

from hypothesis import given, example
from hypothesis.strategies import decimals

ion_decimals = decimals(allow_nan=False, allow_infinity=False)

@given(ion_decimals)
@example(Decimal('1005826281919371473355538432.5'))
def test_decimal(value):
    recovered_value = Decimal(ion.loads(ion.dumps(value)))
    assert recovered_value == value
```

But it uses Hypothesis for generating test cases and I don't know if you want to add this as a development dependency. If anyone can add tests to this PR, feel free to do it.

### DIFF
--- a/amazon/ion/reader_binary.py
+++ b/amazon/ion/reader_binary.py
@@ -198,7 +198,7 @@ def _parse_decimal(buf):
         with localcontext() as context:
             # Adjusting precision for taking into account arbitrarily
             # large/small numbers
-            context.prec = len(str(coefficient)) + abs(exponent)
+            context.prec = len(str(coefficient))
             value = Decimal(coefficient).scaleb(exponent)
 
     return value
--- a/amazon/ion/reader_binary.py
+++ b/amazon/ion/reader_binary.py
@@ -20,7 +20,7 @@ from __future__ import print_function
 import six
 
 from datetime import timedelta
-from decimal import Decimal
+from decimal import Decimal, localcontext
 from functools import partial
 from io import BytesIO
 from struct import unpack
@@ -186,6 +186,7 @@ def _parse_signed_int_components(buf):
 
 def _parse_decimal(buf):
     """Parses the remainder of a file-like object as a decimal."""
+    from decimal import localcontext
     exponent = _parse_var_int(buf, signed=True)
     sign_bit, coefficient = _parse_signed_int_components(buf)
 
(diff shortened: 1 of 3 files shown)


amazon/ion/reader_binary.py
```diff
@@ -194,7 +195,11 @@ def _parse_decimal(buf):
         value = Decimal((sign_bit, (0,), exponent))
     else:
         coefficient *= sign_bit and -1 or 1
-        value = Decimal(coefficient).scaleb(exponent)
+        with localcontext() as context:
+            # Adjusting precision for taking into account arbitrarily
+            # large/small numbers
+            context.prec = len(str(coefficient)) + abs(exponent)
```
pbcornell, line 201: Let's remove `+ abs(exponent)` from this expression.  Setting precision to the length of the coefficient should be sufficient.

amazon/ion/writer_binary_raw.py
```diff
@@ -185,7 +185,11 @@ def _serialize_decimal(ion_event):
     value = ion_event.value
     validate_scalar_value(value, Decimal)
     sign, digits, exponent = value.as_tuple()
-    coefficient = int(value.scaleb(-exponent).to_integral_value())
+    with localcontext() as context:
+        # Adjusting precision for taking into account arbitrarily large/small
+        # numbers
+        context.prec = len(digits) + abs(exponent)
```
pbcornell, line 191: Similarly, this can simply be `context.prec = len(digits)`

pbcornell, line 191: Similarly, I believe this can be simplified to `context.prec = len(digits)`

amazon/ion/reader_binary.py
oscar26: Changes made.

amazon/ion/writer_binary_raw.py
oscar26: Changes made.

oscar26:
Changes incorporated and test added.

Also added you as collaborator to the repo I've forked in any case you want to make further changes.

Thanks for your time!

Issue #107, opened 2019-10-14
Support Ion to JSON


clacrone-github:
Often we have scenarios where input comes as JSON, converted to ION to do some business requirements, append data, etc, and then needs to be converted back to JSON. IonJava and others support the capability to translate ION to JSON. Add this extension for Python please.

tgregg:
Note to anyone considering implementing this feature: the JSON down-conversion rules that should be followed are listed in the cookbook: http://amzn.github.io/ion-docs/guides/cookbook.html#down-converting-to-json

PR #135, opened 2020-11-19
Add Ion-JSON encoder

simonz-bq:
Implements #107 

### DIFF
--- a/amazon/ion/encoder.py
+++ b/amazon/ion/encoder.py
@@ -36,8 +36,8 @@ class IonEncoder(BaseEncoder):
 
     def isinstance(self, obj, cls):
         if (not hasattr(sys, "pypy_version_info")) and \
-                isinstance(obj, (IonPyList, IonPyNull, IonPyBool, IonPyInt, IonPyFloat, IonPyDecimal, IonPyTimestamp,
-                                 IonPyText, IonPyBytes, IonPySymbol)):
+                isinstance(obj, (IonPyList, IonPyDict, IonPyNull, IonPyBool, IonPyInt, IonPyFloat, IonPyDecimal,
+                                 IonPyTimestamp, IonPyText, IonPyBytes, IonPySymbol)):
             return False
         return isinstance(obj, cls)
 
--- a/amazon/ion/encoder.py
+++ b/amazon/ion/encoder.py
@@ -60,4 +60,4 @@ class IonEncoder(JSONExtendedEncoder):
                 return None
             return float(o)
         else:
-            super().default(o)
+            super(IonEncoder, self).default(o)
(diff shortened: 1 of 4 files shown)


amazon/ion/encoder.py
```diff
@@ -0,0 +1,63 @@
...
+            # S-expressions are printed as JSON array
+            return list(map(self.default, o))
+        elif isinstance(o, IonPyDict) and o.ion_type == IonType.STRUCT:
+            return {key: self.default(o[key]) for key in o.keys()}
+        elif isinstance(o, IonPyNull):
+            return None
+        elif isinstance(o, IonPyBytes) and o.ion_type == IonType.BLOB:
+            return standard_b64encode(o).decode("utf-8")
+        elif isinstance(o, IonPyBytes) and o.ion_type == IonType.CLOB:
+            return o.decode("utf-8")
```
tgregg, line 48: You may have to write a custom function (or reuse one from elsewhere in the library, if possible). The JSON-downconversion rules say:

> Clobs are ASCII-encoded for characters between 32 (0x20) and 126 (0x7e), inclusive. Characters from 0 (0x00) to 31 (0x1f) and from 127 (0x7f) to 255 (0xff) are escaped as Unicode code points U+00XX (e.g. 0x7f is U+007f, represented by \u007f in JSON).

So, basically, printable ASCII characters are represented by those ASCII characters in the string, while non-printable characters are replaced with unicode escapes.

tests/test_json_encoder.py
```diff
@@ -0,0 +1,187 @@
...
+from amazon.ion.simple_types import IonPyList, IonPyDict, IonPyNull, IonPyBool, IonPyInt, IonPyFloat, IonPyDecimal, \
+    IonPyTimestamp, IonPyText, IonPyBytes, IonPySymbol
+from amazon.ion.symbols import SymbolToken
+from amazon.ion.simpleion import dumps, loads
+import datetime
+from decimal import Decimal
+import json
+
+
+def test_null():
```
tgregg, line 26: Please add some tests for typed nulls.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+jsonconversion==0.2.12
```
tgregg, line 1: This isn't part of the built-in JSON library?

tgregg (commented):
I like it.

tgregg:
The tests failed on Python < 3.6 and pypy. Can you look into that? https://github.com/amzn/ion-python/pull/135/checks?check_run_id=1421663403

tgregg:
Since we're not supporting pypy for the IonEncoder, we should add a check for pypy at the module level and throw a helpful error if it's being used. In the tests, we should add a check that makes sure the error is thrown when pypy is used.

We should rename the module `IonToJsonEncoder` to make it more clear what it is for. We should also add a module-level docstring that explains the purpose of the module and demonstrates how to use it. The pypy restriction should be called out in this docstring.

And we still need some typed null tests.

amazon/ion/encoder.py
simonz-bq: Done. The chr(byte) function was found to do this conversion for us and is now used.

tests/test_json_encoder.py
simonz-bq: Done.

requirements.txt
simonz-bq: This library uses the built-in JSON library and wraps the class logic easily so that we can overwrite it with the logic we need.

tgregg (approved):

Issue #137, opened 2021-01-15
Why `ion_type` needs to be specified separately?


yosoufe:
Hi,

I would like to know why in the following example the `ion_type` needs to be specified separately?

```py
from amazon.ion import simpleion as ion
from collections import OrderedDict
ion_sym = ion.IonPySymbol('SomeSymbol', None)
ion_sym.ion_type = ion.IonType.SYMBOL  # why this line cannot be inside the constructor of IonPy*?

ion_dict = ion.IonPyDict(OrderedDict([
    ("key1", 1),
    ("key2", 2),
    ("someSymbol", ion_sym),
]))
ion_dict.ion_type = ion.IonType.STRUCT # why this line cannot be inside the constructor of IonPy*?
ion_bytes = ion.dumps(ion_dict, binary=True)
```

Every time I create any `IonPy*` object I need to specify the `ion_type` as well. As far as I understand there is a 1 to 1 mapping between each `IonType` and `IonPy*` object. So that could be specified inside the constructor of `IonPy*` class. 

Currently if I do not specify `ion_type` I will get error messages during `ion.dumps`. And when my ion struct is large and nested, it is difficult to find were I missed specifying `ion_type` and also the error message does not give me any hint where I missed it. I have to manually check the complete code, to find where I missed it.

Thanks

cheqianh:
Hi, different ion_types dump different ion-formatted stream.

For example, 
**IonPyNull** can be `null.bool`, `null.int`... by specifying the ion_type,  
**IonPyText('SomeSymbol')** can be either `SomeSymbol` if its ion_type is `IonType.SYMBOL` or `"SomeSymbol"` if its ion_type is `IonType.STRING`, etc.

Refer to [conversion table](https://github.com/amzn/ion-python/blob/a14882559e77dbac74ade54d4a5cf28a0cf01a1d/amazon/ion/simpleion.py#L54) for more details.

yosoufe:
Ok. Thanks. So it is not 1:1 as I thought. Is it possible to generate a better error message to make it easier to find where I am missing the `ion_type`? for example now this code
```.py
from amazon.ion import simpleion as ion
from collections import OrderedDict
ion_sym = ion.IonPySymbol('SomeSymbol', None)

ion_dict = ion.IonPyDict(OrderedDict([
    ("key1", 1),
    ("key2", 2),
    ("someSymbol", ion_sym),
]))
ion_dict.ion_type = ion.IonType.STRUCT
[...]

PR #138, opened 2021-01-30
Issue 137, have a detail error message for missing ion_type.

cheqianh:
### Description:
As mentioned in #137, this PR is trying to generate a detail error message for missing ion_type.

It would be easier to navigate the missing ion_type object by giving object value and its depth. In addition, if the object is inside a STRUCT, the `filed` will be the key instead of None.

&nbsp;
### Example output: 
**1. Missing `ion_sym`'s ion_type** 
```.py
from amazon.ion import simpleion as ion

ion_sym = ion.IonPySymbol('SomeSymbol', None)
# ion_sym.ion_type = ion.IonType.SYMBOL                 comment this line to produce error

ion_bytes = ion.dumps(ion_sym, binary=True)
```
**Error message:**
```.py
amazon.ion.exceptions.IonException: ion_type 'None' is not allowed in value: "IonPySymbol(text='SomeSymbol', sid=None, location=None)", depth: 0, field: None
```

**2. Missing `ion_sym`'s ion_type (inside a struct)**
```.py
from amazon.ion import simpleion as ion
from collections import OrderedDict

ion_sym = ion.IonPySymbol('SomeSymbol', None)
# ion_sym.ion_type = ion.IonType.SYMBOL                  comment this line to produce error

ion_dict = ion.IonPyDict(OrderedDict([
    ("key1", 1),
    ("key2", 2),
    ("someSymbol", ion_sym),
]))
ion_dict.ion_type = ion.IonType.STRUCT 
ion_bytes = ion.dumps(ion_dict, binary=True)
```
**Error message:** 
```.py
amazon.ion.exceptions.IonException: ion_type 'None' is not allowed in value: "IonPySymbol(text='SomeSymbol', sid=None, location=None)", depth: 1, field: someSymbol
```

**3. Missing `ion_dict`'s ion_type**
```.py
from amazon.ion import simpleion as ion
from collections import OrderedDict

ion_sym = ion.IonPySymbol('SomeSymbol', None)
ion_sym.ion_type = ion.IonType.SYMBOL

ion_dict = ion.IonPyDict(OrderedDict([
    ("key1", 1),
    ("key2", 2),
    ("someSymbol", ion_sym),
]))
# ion_dict.ion_type = ion.IonType.STRUCT                comment this line to produce error
ion_bytes = ion.dumps(ion_dict, binary=True)
```
**Error message:**
```.py
amazon.ion.exceptions.IonException: ion_type 'None' is not allowed in value: "<amazon.ion.simple_types.IonPyDict object at 0x10a232b80>", depth: 0, field: None
```

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -210,7 +210,7 @@ def _dump(obj, writer, from_type, field=None, in_struct=False, depth=0):
         ion_type = _ion_type(obj, from_type)
         ion_nature = False
     if ion_type is None:
-        raise IonException('ion_type \'None\' is not allowed in value: \"%s\", depth: %d, field: %s' % (repr(obj), depth, field))
+        raise IonException('Value must have a non-None ion_type: %s, depth: %d, field: %s' % (repr(obj), depth, field))
     if not null and ion_type.is_container:
         if ion_nature:
             event = obj.to_event(IonEventType.CONTAINER_START, field_name=field, in_struct=in_struct, depth=depth)
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -209,6 +209,9 @@ def _dump(obj, writer, from_type, field=None, in_struct=False, depth=0):
     except AttributeError:
         ion_type = _ion_type(obj, from_type)
         ion_nature = False
+    if ion_type is None:
+        obj_str = repr(obj) if field is None else '%s: %s' % (field, repr(obj))
+        raise IonException('ion_type \'None\' is not allowed in value: \"%s\", depth: %d' % (obj_str, depth))
     if not null and ion_type.is_container:
         if ion_nature:
             event = obj.to_event(IonEventType.CONTAINER_START, field_name=field, in_struct=in_struct, depth=depth)
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -210,8 +210,7 @@ def _dump(obj, writer, from_type, field=None, in_struct=False, depth=0):
         ion_type = _ion_type(obj, from_type)
         ion_nature = False
     if ion_type is None:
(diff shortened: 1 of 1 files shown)


amazon/ion/simpleion.py
```diff
@@ -209,6 +209,8 @@ def _dump(obj, writer, from_type, field=None, in_struct=False, depth=0):
     except AttributeError:
         ion_type = _ion_type(obj, from_type)
         ion_nature = False
+    if ion_type is None:
+        raise IonException('ion_type \'None\' is not allowed in value: \"%s\", depth: %d, field: %s' % (repr(obj), depth, field))
```
tgregg, line 213: I would say something like "Value must have a non-None ion_type"

tgregg (commented):
I think adding an error message is fine, but I also think that it should be much harder to construct an Ion value type without its ion_type. In your example, you've used IonPySymbol and IonPyDict constructors explicitly. These constructors implicitly know what the ion_type should be for that value (SYBMOL and STRUCT, respectively), so it should set that type. Also, [`from_value`](https://github.com/amzn/ion-python/blob/master/amazon/ion/simple_types.py#L90) should be the way most commonly used to construct these types, and that requires ion_type to be specified.

cheqianh:
The current message looks like below:

```.py
amazon.ion.exceptions.IonException: Value must have a non-None ion_type: IonPySymbol(text='sym', sid=12, location=None), depth: 0, field: None
```

Yeah, IonType is important to create a value. Even though fromValue is the most common way to create a value, let me know If you think there is anything else I can do for this situation.

tgregg (approved):
This PR can be merged. Would you mind opening an issue to correctly set the ion_type when the simple_types constructors are invoked explicitly?

cheqianh:
> This PR can be merged. Would you mind opening an issue to correctly set the ion_type when the simple_types constructors are invoked explicitly?

Sure!

Issue #140, opened 2021-02-23
Empty String Key in Dict Fails Encoding


bryanhelmig:
Reproduction case:

```python
import amazon.ion.simpleion as ion
ion.dumps({'': ''})
```

Gives you:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/simpleion.py", line 283, in dumps
    dump(obj, ion_buffer, imports=imports, sequence_as_stream=sequence_as_stream, binary=binary, skipkeys=skipkeys,
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/simpleion.py", line 158, in dump
    _dump(obj, writer, from_type)
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/simpleion.py", line 220, in _dump
    _dump(val, writer, from_type, field, in_struct=True, depth=depth+1)
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/simpleion.py", line 231, in _dump
    writer.send(event)
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/writer.py", line 150, in blocking_writer
    for result_event in _drain(writer, ion_event):
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/writer.py", line 129, in _drain
    result_event = writer.send(ion_event)
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/writer.py", line 111, in writer_trampoline
    trans = trans.delegate.send(Transition(ion_event, trans.delegate))
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/writer_binary.py", line 214, in _managed_binary_writer_coroutine
    write_event = value_writer.send(ion_event)
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/writer.py", line 111, in writer_trampoline
    trans = trans.delegate.send(Transition(ion_event, trans.delegate))
  File "/code/ion-is-cool/.venv/lib/python3.8/site-packages/amazon/ion/writer_binary_raw.py", line 353, in _raw_writer_coroutine
    _write_varuint(sid_buffer, ion_event.field_name.sid)  # Write the field name's symbol ID.
AttributeError: 'str' object has no attribute 'sid'
[...]

PR #141, opened 2021-02-24
Added support of empty string field_name.

cheqianh:
### Description:

This PR worked on issue #140.

The binary writer doesn't include the case that field_name is an empty string.

&nbsp;
### TEST:
**Test code:**

```.py
from amazon.ion.simpleion import dumps

obj = {'': ''}
print(dumps(obj, binary=True))
```

**Output:** 
```
b'\xe0\x01\x00\xea\xe7\x81\x83\xde\x83\x87\xb1\x80\xde\x82\x8a\x80'
```
which is: `$ion    ion_symbol_table::{  symbols:[""]  }     {"" :  ""} `

&nbsp;
**Unit test**:

Screen Shot 2021-02-23 at 7 46 11 PM

### DIFF
--- a/amazon/ion/writer_binary.py
+++ b/amazon/ion/writer_binary.py
@@ -168,7 +168,7 @@ def _managed_binary_writer_coroutine(imports):
     def intern_symbols(event):
         field_name = event.field_name
         annotations = event.annotations
-        if field_name:
+        if field_name or field_name == '':
             event = event.derive_field_name(intern_symbol(field_name))
         if annotations:
             event = event.derive_annotations(
--- a/tests/test_simpleion.py
+++ b/tests/test_simpleion.py
@@ -130,6 +130,18 @@ _SIMPLE_CONTAINER_MAP = {
             _Expected(b'\xD0', b'{}')
         ),
         (
+            [{u'': u''}, ],
+            _Expected(
+                bytes_of([
+                    0xDE,  # The lower nibble may vary. It does not indicate actual length unless it's 0.
+                    VARUINT_END_BYTE | 2,  # Field name 10 and value 0 each fit in 1 byte.
+                    VARUINT_END_BYTE | 10,
+                    0x80    # Empty string
+                ]),
+                b"{'':\"\"}"
+            )
+        ),
+        (
             [{u'foo': 0}, ],
             _Expected(
                 bytes_of([


tgregg:
Two questions:
1. There are test files in `ion-tests` that exercise empty field names. Why were they not failing? Are they on the skip list?
2. Were our pypy Travis builds failing before previously?

cheqianh:
1. We had specific tests that cover empty filed name. The reason causing this issue is because of the difference ion_type(or if they are ion nature).
&nbsp;
**When we are running the unit test**: We load unicode (from vector files) to a `IonPyDict` object which is an ion nature object. Then when we dump the object, we go through this [branch](https://github.com/amzn/ion-python/blob/f8c61fbf1134071da0151cba2c045cb9e390ff7a/amazon/ion/simpleion.py#L216) to create an ionEvent by `to_event` method that will convert field name to a symbol token. 
&nbsp;
**When we dump a python object {'': ''} directly**: Since it's not an ion nature, it will go through this [branch](https://github.com/amzn/ion-python/blob/f8c61fbf1134071da0151cba2c045cb9e390ff7a/amazon/ion/simpleion.py#L218) calling IonEvent() constructor to generate an ionEvent. IonEvent supports both unicode or symbol token as filed name so "" will be treated as unicode first and later on we will call the `derive_field_name` method [here](https://github.com/amzn/ion-python/blob/f8c61fbf1134071da0151cba2c045cb9e390ff7a/amazon/ion/writer_binary.py#L172) to convert the filed name from unicode to symbol token. 
&nbsp;
The issue happenes when we are going to call the `derive_field_name` method [here](https://github.com/amzn/ion-python/blob/f8c61fbf1134071da0151cba2c045cb9e390ff7a/amazon/ion/writer_binary.py#L171). we don't include the case that field name can be an empty string `""`  which throws an error later.

2. I'm not very sure about the pypy builds failing, I noticed that some of our previous PRs faced the same issue.

tgregg:
1. Okay, in that case can you please add a unit test that targets this code path?
2. Got it, if you have a few minutes to look into it after this PR and at least create an issue for fixing it, that would be great.

cheqianh:
1. Added a new unit test covering this case. Before this PR, the new test fails due to the empty string field issue, and passed after this PR change.
**before**: Screen Shot 2021-03-03 at 7 43 04 PM
**now**: Screen Shot 2021-03-03 at 7 41 33 PM

2. Sure, I'll see if this commit fails again to ensure it's not a random issue.

tgregg (approved):

PR #144, opened 2021-03-19
Adds `__repr__` for ionPyDict.

cheqianh:
### Description:

Worked on Issue#86 for a friendlier `repr()` implementation of IonPyDict. 
This is a first version of` __repr__` method that prints value/key pair. 

Added [MultiMap](https://github.com/amzn/ion-python/blob/8bf102218ae2bacfa7f6ddbc61dce4d1e1dd8374/amazon/ion/core.py#L598)'s `__repr__` method, it iterates each value and recursively call other ionPyValue's `__repr__()` if needed.

### Test:
Sample code:
```.py
import io

myobj =myobj = {'StrKey': 'StrValue', 'IntKey': 123, 'boolKey': False, 'listKey': [1], 'structKey': {'true': 'hello ion'}} # <---- output
fp = io.BytesIO()
ion.dump(myobj, fp)
ion_dict = ion.load(io.BytesIO(fp.getvalue()))
print(ion_dict)
```

**Before**:
`<amazon.ion.simple_types._ion_type_for.<locals>.IonPyValueType object at 0x100947cc0>`

**Now:**
`{'StrKey': 'StrValue', 'IntKey': 123, 'boolKey': 0, 'listKey': [1], 'structKey': {'true': 'hello ion'}}`


cheqianh:
Trying to figure out why travis failed.

cheqianh:
I'm going to set this PR as draft mode and test a little bit until travis build passes.

cheqianh:
The build failed due to the python version 2's __repr__ method doesn't support unicode return value. It is fixed in the later version (3+).

__str__ and __repr__ have a similar usage but I used __str__ method eventually since:
> Well, the __str__ function is supposed to return a human-readable format, which is good for logging or to display some information about the object. Whereas, the __repr__ function is supposed to return an “official” string representation of the object, which can be used to construct the object again.

If there is no other concerns for this PR, the next step is writing some doc.

amazon/ion/core.py
```diff
@@ -630,6 +633,15 @@ def __iter__(self):
         for key in six.iterkeys(self.__store):
             yield key
 
+    if _PY_VERSION >= 3:
```
tgregg, line 636: Instead of only defining `__str__` for Python 3+, we should try to make it work for Python 2 as well. `six` should be able to provide an abstraction that you can use.

amazon/ion/core.py
```diff
@@ -630,6 +633,15 @@ def __iter__(self):
...
+        def __str__(self):
+            str_repr = '{'
+            for key in six.iterkeys(self.__store):
+                str_repr += '\'%s\': %s, ' % (key, self.__store[key][0].__repr__())
```
tgregg, line 640: Call `repr(self.__store[key][0])` instead of `self.__store[key][0].__repr__()`. Or, should this be `str(self.__store[key][0])`?

amazon/ion/core.py
```diff
@@ -630,6 +633,15 @@ def __iter__(self):
...
+            str_repr = str_repr[:str_repr.__len__() - 2] + '}'
```
tgregg, line 641: Call `len(str_repr)` instead of `str_repr.__len__()`.

cheqianh:
Okay build passed using six module.

tgregg:
Looks good. I would suggest also supplying an implementation of `__repr__` that just returns the `str` representation. It's fine if they're the same. There may be users who use both.

cheqianh:
going to test python 2/3 for __repr__

cheqianh:
I added __repr__ = __str__ but it fails travis with <2 version. I tried couple ways such as using .encode() or six.ensure_string but neither worked. 

In case I missed anything, Do you have any idea about this before I do more research?

cheqianh:
Return six.ensure_binary for py2 solves travis failed issue.

tgregg (approved):
Looks good, please squash and merge.

PR #149, opened 2021-04-22
C extension (speeds up simpleion dump/load)

cheqianh:
# Description:
PR of C extension.  (both coding and configuration parts).

Distribution part may be changed but I'll create a separate PRs for them as needed. 

I added comment for myself from the feedback last time. (e.g. not use force-python-impl, check c skip list etc.)

# Suggested Review Order

 C extension consists of three parts: `simpleion dump/load`, `c extension (e.g. for mac it's ionc.so)` and `ion-c binaries`. We will go through each of them below. 

### C extension --- `setup.py`: ###  
**Highlights:**
We add c extension information [here](https://github.com/amzn/ion-python/pull/149/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R26), setup.py will build c extension during the build process. we can also manually build extension by `python setup.py build_ext`.

&nbsp;

### Ion-c binaries --- `install.py`: ###
Then we look into `install.py` where we prepare ion-c binaries and headers for us. it starts from [here](https://github.com/amzn/ion-python/pull/149/files#diff-a5fa297e66b71ac1a8685caa5c37577774be2c36024ecda7abeb70bb50214d0fR173).

**Highlights:**
1. The high level goal of install.py is: 1. checkout ion-c, 2. build ion-c binaries, 3. move required output for `ionc` and `decNumber` into `ion-c-build`. 4. copy `ion-c-build` to `amazon/ion` for distribution.
2. We currently support Mac, Windows and Linux. There are still some issues need to be improve. I added comment regarding these issues. (e.g. in windows, Check if VS is available in virtual machine; In linux, the wheel is initialized with an unexpected name, but simply rename it solves issue, need more research about this. etc.)

&nbsp;

### Python --- `simpleion.py`: ###
Now We look into simpleion.py, how we connect ion-python and c extension.

**Highlights:**
1. In order to **not** use `force_python_impl`, I rename the previous `dump/load` to `dump_original/load_original`. New `dump` method wraps `dump_original` and `dump_extension` (c extension dump API). So does `load`.

&nbsp;

### Distribute all required files --- `MANIFEST.in`: ###
After preparing all three parts, the next step is letting PYPI know what files we are going to distribute.

We added some required files into [MANIFEST.in](https://github.com/amzn/ion-python/pull/149/files#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003R6) for distribution: ion-c binaries in `ion-c-build`, `_ioncmodule.h` header file for c module and installing script `install.py`. Now when we distribute ion-python, both c extension and ion-c will be included. 

&nbsp;
&nbsp;
### ion-c module --- `ioncmodule.c and _ioncmodule.h` ###
For `dump (write)`, it starts from [here](https://github.com/amzn/ion-python/pull/149/files#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR744), this method initialize required variable and parse the python argument to c types. It will eventually call [ionc_write_value](https://github.com/amzn/ion-python/pull/149/files#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR483) to write value. `ionc_write_value` will check what type the value is and use the corresponding ion-c api to write them.

Likely for `load`, it starts from [here](https://github.com/amzn/ion-python/pull/149/files#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR1191) and eventually will call [ionc_read_value](https://github.com/amzn/ion-python/pull/149/files#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR939) to read files.

&nbsp;
#### After knowing how the c module looks like, we move on to unit tests: ####
`tests/test_cookbook.py` 
`tests/test_simpleion.py`
`tests/tests/test_vectors.py`
Refer to comments for more details.

&nbsp;
#### Last file `.github/workflows/python-publish.yml` is minor, ####
Don't need to review anymore, going to find another way to automate distribution.

I'll temporary keep this script until we find a way to distribute.

## ##

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -1,0 +1,1369 @@
+#include <Python.h>
+#include "datetime.h"
+#include "_ioncmodule.h"
+
+#define cRETURN RETURN(__location_name__, __line__, __count__++, err)
+
+#define YEAR_PRECISION 0
+#define MONTH_PRECISION 1
+#define DAY_PRECISION 2
+#define MINUTE_PRECISION 3
+#define SECOND_PRECISION 4
+
+#define MICROSECOND_DIGITS 6
+
+#define ERR_MSG_MAX_LEN 100
+#define FIELD_NAME_MAX_LEN 100
+#define DEC_NUMBER_MAX_LEN 300
+
+static char _err_msg[ERR_MSG_MAX_LEN];
+
+#define _FAILWITHMSG(x, msg) { err = x; snprintf(_err_msg, ERR_MSG_MAX_LEN, msg); goto fail; }
+
+// Python 2/3 compatibility
+#if PY_MAJOR_VERSION >= 3
+    #define IONC_BYTES_FORMAT "y#"
+    #define IONC_READ_ARGS_FORMAT "OOO"
+    #define PyInt_AsSsize_t PyLong_AsSsize_t
+    #define PyInt_AsLong PyLong_AsLong
+    #define PyInt_FromLong PyLong_FromLong
+    #define PyString_AsStringAndSize PyBytes_AsStringAndSize
+    #define PyString_Check PyUnicode_Check
+    #define PyString_FromStringAndSize PyUnicode_FromStringAndSize
+    #define PyString_FromString PyUnicode_FromString
+    #define PyInt_Check PyLong_Check
+#else
+    #define IONC_BYTES_FORMAT "s#"
+    #define IONC_READ_ARGS_FORMAT "OOO"
+#endif
+
+#if PY_VERSION_HEX < 0x02070000
+    #define offset_seconds(x) offset_seconds_26(x)
+#endif
+
+static PyObject* _decimal_module;
+static PyObject* _decimal_constructor;
+static PyObject* _py_timestamp_constructor;
+static PyObject* _simpletypes_module;
+static PyObject* _ionpynull_cls;
+static PyObject* _ionpynull_fromvalue;
+static PyObject* _ionpybool_cls;
+static PyObject* _ionpybool_fromvalue;
+static PyObject* _ionpyint_cls;
+static PyObject* _ionpyint_fromvalue;
+static PyObject* _ionpyfloat_cls;
+static PyObject* _ionpyfloat_fromvalue;
+static PyObject* _ionpydecimal_cls;
+static PyObject* _ionpydecimal_fromvalue;
+static PyObject* _ionpytimestamp_cls;
+static PyObject* _ionpytimestamp_fromvalue;
+static PyObject* _ionpytext_cls;
+static PyObject* _ionpytext_fromvalue;
+static PyObject* _ionpysymbol_cls;
+static PyObject* _ionpysymbol_fromvalue;
+static PyObject* _ionpybytes_cls;
+static PyObject* _ionpybytes_fromvalue;
+static PyObject* _ionpylist_cls;
+static PyObject* _ionpylist_fromvalue;
+static PyObject* _ionpydict_cls;
+static PyObject* _ionpydict_fromvalue;
+static PyObject* _ion_core_module;
+static PyObject* _py_ion_type;
+static PyObject* py_ion_type_table[14];
+static int  c_ion_type_table[14];
+static PyObject* _py_timestamp_precision;
+static PyObject* py_ion_timestamp_precision_table[7];
+static PyObject* _ion_symbols_module;
+static PyObject* _py_symboltoken_constructor;
+static PyObject* _exception_module;
+static PyObject* _ion_exception_cls;
+static decContext dec_context;
+
+
+/******************************************************************************
+*       helper functions                                                      *
+******************************************************************************/
+
+/*
+ *  Gets an attribute as an int. NOTE: defaults to 0 if the attribute is None.
+ *
+ *  Args:
+ *      obj: An object whose attribute will be returned
+ *      attr_name: An attribute of the object
+ *
+ *  Returns:
+ *      An attribute as an int
+ */
+static int int_attr_by_name(PyObject* obj, char* attr_name) {
+    PyObject* py_int = PyObject_GetAttrString(obj, attr_name);
+    int c_int = 0;
+    if (py_int != Py_None) {
+        c_int = (int)PyInt_AsSsize_t(py_int);
+    }
+    Py_DECREF(py_int);
+    return c_int;
+}
+
+// TODO compare performance of these offset_seconds* methods. The _26 version will work with all versions, so if it is
+// as fast, should be used for all.
+static int offset_seconds_26(PyObject* timedelta) {
+    long microseconds = int_attr_by_name(timedelta, "microseconds");
+    long seconds_microseconds = (long)int_attr_by_name(timedelta, "seconds") * 1000000;
+    long days_microseconds = (long)int_attr_by_name(timedelta, "days") * 24 * 3600 * 1000000;
+    return (microseconds + seconds_microseconds + days_microseconds) / 1000000;
+}
+
+static int offset_seconds(PyObject* timedelta) {
+    PyObject* py_seconds = PyObject_CallMethod(timedelta, "total_seconds", NULL);
+    PyObject* py_seconds_int = PyObject_CallMethod(py_seconds, "__int__", NULL);
+    int seconds = (int)PyInt_AsSsize_t(py_seconds_int);
+    Py_DECREF(py_seconds);
+    Py_DECREF(py_seconds_int);
+    return seconds;
+}
+
+/*
+ *  Returns the ion type of an object as an int
+ *
+ *  Args:
+ *      obj: An object whose type will be returned
+ *
+ *  Returns:
+ *      An int in 'c_ion_type_table' representing an ion type
+ */
+static int ion_type_from_py(PyObject* obj) {
+    PyObject* ion_type = NULL;
+    if (PyObject_HasAttrString(obj, "ion_type")) {
+        ion_type = PyObject_GetAttrString(obj, "ion_type");
+    }
+    if (ion_type == NULL) return tid_none_INT;
+    int c_type = c_ion_type_table[PyInt_AsSsize_t(ion_type)];
+    Py_DECREF(ion_type);
+    return c_type;
+}
+
+/*
+ *  Gets a C string from a python string
+ *
+ *  Args:
+ *      str:  A python string that needs to be converted
+ *      out:  A C string converted from 'str'
+ *      len_out:  Length of 'out'
+ */
+static iERR c_string_from_py(PyObject* str, char** out, Py_ssize_t* len_out) {
+    iENTER;
+#if PY_MAJOR_VERSION >= 3
+    *out = PyUnicode_AsUTF8AndSize(str, len_out);
+#else
+    PyObject *utf8_str;
+    if (PyUnicode_Check(str)) {
+        utf8_str = PyUnicode_AsUTF8String(str);
+    }
+    else {
+        utf8_str = PyString_AsEncodedObject(str, "utf-8", "strict");
+    }
+    if (!utf8_str) {
+        _FAILWITHMSG(IERR_INVALID_ARG, "Python 2 fails to convert python string to utf8 string.");
+    }
+    PyString_AsStringAndSize(utf8_str, out, len_out);
+    Py_DECREF(utf8_str);
+#endif
+    iRETURN;
+}
+
+/*
+ *  Gets an ION_STRING from a python string
+ *
+ *  Args:
+ *      str:  A python string that needs to be converted
+ *      out:  An ION_STRING converted from 'str'
+ */
+static iERR ion_string_from_py(PyObject* str, ION_STRING* out) {
+    iENTER;
+    char* c_str = NULL;
+    Py_ssize_t c_str_len;
+    IONCHECK(c_string_from_py(str, &c_str, &c_str_len));
+    ION_STRING_INIT(out);
+    ion_string_assign_cstr(out, c_str, c_str_len);
+    iRETURN;
+}
+
+/*
+ *  Builds a python string using an ION_STRING
+ *
+ *  Args:
+ *      string_value:  An ION_STRING that needs to be converted
+ *
+ *  Returns:
+ *      A python string
+ */
+static PyObject* ion_build_py_string(ION_STRING* string_value) {
+    // TODO Test non-ASCII compatibility.
+    // NOTE: this does a copy, which is good.
+    if (!string_value->value) return Py_None;
+    return PyUnicode_FromStringAndSize((char*)(string_value->value), string_value->length);
+}
+
+/*
+ *  Adds an element to a List or struct
+ *
+ *  Args:
+ *      pyContainer:  A container that the element is added to
+ *      element:  The element to be added to the container
+ *      in_struct:  if the current state is in a struct
+ *      field_name:  The field name of the element if it is inside a struct
+ */
+static void ionc_add_to_container(PyObject* pyContainer, PyObject* element, BOOL in_struct, ION_STRING* field_name) {
+    if (in_struct) {
+        PyObject_CallMethodObjArgs(
+            pyContainer,
+            PyString_FromString("add_item"),
+            ion_build_py_string(field_name),
+            (PyObject*)element,
+            NULL
+        );
+    }
+    else {
+        PyList_Append(pyContainer, (PyObject*)element);
+    }
+    Py_XDECREF(element);
+}
+
+/*
+ *  Converts an ion decimal string to a python-decimal-accept string. NOTE: ion spec uses 'd' in a decimal number
+ *  while python decimal object accepts 'e'
+ *
+ *  Args:
+ *      dec_str:  A C string representing a decimal number
+ *
+ */
+static void c_decstr_to_py_decstr(char* dec_str) {
+    for (int i = 0; i < strlen(dec_str); i++) {
+        if (dec_str[i] == 'd' || dec_str[i] == 'D') {
+            dec_str[i] = 'e';
+        }
+    }
+}
+
+/*
+ *  Returns a python symbol token using an ION_STRING
+ *
+ *  Args:
+ *      string_value:  An ION_STRING that needs to be converted
+ *
+ *  Returns:
+ *      A python symbol token
+ */
+static PyObject* ion_string_to_py_symboltoken(ION_STRING* string_value) {
+    PyObject* py_string_value;
+    PyObject* py_sid;
+    if (string_value->value) {
+        py_string_value = ion_build_py_string(string_value);
+        py_sid = Py_None;
+    }
+    else {
+        py_string_value = Py_None;
+        py_sid = PyLong_FromLong(0);
+    }
+    return PyObject_CallFunctionObjArgs(
+        _py_symboltoken_constructor,
+        py_string_value,
+        py_sid,
+        NULL
+    );
+}
+
+
+/******************************************************************************
+*       Write/Dump APIs                                                       *
+******************************************************************************/
+
+
+/*
+ *  Writes a symbol token. NOTE: It can be either a value or an annotation
+ *
+ *  Args:
+ *      writer:  An ion writer
+ *      symboltoken: A python symbol token
+ *      is_value: Writes a symbol token value if is_value is TRUE, otherwise writes an annotation
+ *
+ */
+static iERR ionc_write_symboltoken(hWRITER writer, PyObject* symboltoken, BOOL is_value) {
+    iENTER;
+    PyObject* symbol_text = PyObject_GetAttrString(symboltoken, "text");
+    if (symbol_text == Py_None) {
+        PyObject* py_sid = PyObject_GetAttrString(symboltoken, "sid");
+        SID sid = PyInt_AsSsize_t(py_sid);
+        if (is_value) {
+            err = _ion_writer_write_symbol_id_helper(writer, sid);
+        }
+        else {
+            err = _ion_writer_add_annotation_sid_helper(writer, sid);
+        }
+        Py_DECREF(py_sid);
+    }
+    else {
+        ION_STRING string_value;
+        ion_string_from_py(symbol_text, &string_value);
+        if (is_value) {
+            err = ion_writer_write_symbol(writer, &string_value);
+        }
+        else {
+            err = ion_writer_add_annotation(writer, &string_value);
+        }
+    }
+    Py_DECREF(symbol_text);
+    IONCHECK(err);
+    iRETURN;
+}
+
+/*
+ *  Writes annotations
+ *
+ *  Args:
+ *      writer:  An ion writer
+ *      obj: A sequence of ion python annotations
+ *
+ */
+static iERR ionc_write_annotations(hWRITER writer, PyObject* obj) {
+    iENTER;
+    PyObject* annotations = NULL;
+    if (PyObject_HasAttrString(obj, "ion_annotations")) {
+        annotations = PyObject_GetAttrString(obj, "ion_annotations");
+    }
+
+    if (annotations == NULL || PyObject_Not(annotations)) SUCCEED();
+
+    annotations = PySequence_Fast(annotations, "expected sequence");
+    Py_ssize_t len = PySequence_Size(annotations);
+    Py_ssize_t i;
+
+    for (i = 0; i < len; i++) {
+        PyObject* pyAnnotation = PySequence_Fast_GET_ITEM(annotations, i);
+        Py_INCREF(pyAnnotation);
+        if (PyUnicode_Check(pyAnnotation)) {
+            ION_STRING annotation;
+            ion_string_from_py(pyAnnotation, &annotation);
+            err = ion_writer_add_annotation(writer, &annotation);
+        }
+        else if (PyObject_TypeCheck(pyAnnotation, (PyTypeObject*)_py_symboltoken_constructor)){
+            err = ionc_write_symboltoken(writer, pyAnnotation, /*is_value=*/FALSE);
+        }
+        Py_DECREF(pyAnnotation);
+        if (err) break;
+    }
+    Py_XDECREF(annotations);
+fail:
+    Py_XDECREF(annotations);
+    cRETURN;
+}
+
+/*
+ *  Writes a list or a sexp
+ *
+ *  Args:
+ *      writer:  An ion writer
+ *      sequence: An ion python list or sexp
+ *      tuple_as_sexp: Decides if a tuple is treated as sexp
+ *
+ */
+static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tuple_as_sexp) {
+    iENTER;
+    PyObject* child_obj = NULL;
+    sequence = PySequence_Fast(sequence, "expected sequence");
+    Py_ssize_t len = PySequence_Size(sequence);
+    Py_ssize_t i;
+
+    for (i = 0; i < len; i++) {
+        child_obj = PySequence_Fast_GET_ITEM(sequence, i);
+        Py_INCREF(child_obj);
+
+        IONCHECK(Py_EnterRecursiveCall(" while writing an Ion sequence"));
+        err = ionc_write_value(writer, child_obj, tuple_as_sexp);
+        Py_LeaveRecursiveCall();
+        IONCHECK(err);
+
+        Py_DECREF(child_obj);
+        child_obj = NULL;
+    }
+fail:
+    Py_XDECREF(child_obj);
+    Py_DECREF(sequence);
+    cRETURN;
+}
+
+/*
+ *  Writes a struct
+ *
+ *  Args:
+ *      writer:  An ion writer
+ *      map: An ion python struct
+ *      tuple_as_sexp: Decides if a tuple is treated as sexp
+ *
+ */
+static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_sexp) {
+    iENTER;
+    PyObject * list = PyMapping_Items(map);
+    PyObject * seq = PySequence_Fast(list, "expected a sequence within the map.");
+    PyObject * key = NULL, *val = NULL, *child_obj = NULL;
+    Py_ssize_t len = PySequence_Size(seq);
+    Py_ssize_t i;
+
+    for (i = 0; i < len; i++) {
+        child_obj = PySequence_Fast_GET_ITEM(seq, i);
+        key = PyTuple_GetItem(child_obj, 0);
+        val = PyTuple_GetItem(child_obj, 1);
+        Py_INCREF(child_obj);
+        Py_INCREF(key);
+        Py_INCREF(val);
+
+        if (PyUnicode_Check(key)) {
+            ION_STRING field_name;
+            ion_string_from_py(key, &field_name);
+            IONCHECK(ion_writer_write_field_name(writer, &field_name));
+        }
+        else if (key == Py_None) {
+            // if field_name is None, write symbol $0 instead.
+            IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
+        }
+
+        IONCHECK(Py_EnterRecursiveCall(" while writing an Ion struct"));
+        err = ionc_write_value(writer, val, tuple_as_sexp);
+        Py_LeaveRecursiveCall();
+        IONCHECK(err);
+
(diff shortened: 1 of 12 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+/*
+ *  Writes a struct
+ *
+ *  Args:
+ *      writer:  An ion writer
+ *      map: An ion python struct
+ *      tuple_as_sexp: Decides if a tuple is treated as sexp
+ *
+ */
+static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_sexp) {
```
cheqianh, line 404: Before we use python dict for ion struct which didn't support duplicate keys, I re-write this api.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+            _FAILWITHMSG(IERR_INVALID_ARG, "Found Decimal; expected DECIMAL Ion type.");
+        }
+
+        PyObject* decimal_str = PyObject_CallMethod(obj, "__str__", NULL); // TODO converting every decimal to string is slow.
+        char* decimal_c_str = NULL;
+        Py_ssize_t decimal_c_str_len;
+        IONCHECK(c_string_from_py(decimal_str, &decimal_c_str, &decimal_c_str_len));
+
+        ION_DECIMAL decimal_value;
+        IONCHECK(ion_decimal_from_string(&decimal_value, decimal_c_str, &dec_context));
```
cheqianh, line 569: Replaced previous `decQuadFromString` to `ion_decimal_from_string`.

Refer to [here](https://github.com/cheqianh/ion-python/blob/43bf9243c9b145f8e2baca32624f693bfb3d87ab/amazon/ion/ioncmodule.c#L371) for previous code.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+                _FAILWITHMSG(IERR_INVALID_TIMESTAMP, "Timestamp fractional precision cannot be a positive number.");
+            }
+            fractional_precision = fractional_precision * -1;
+
+            decQuad tmp;
+            decQuadScaleB(&fraction, &fraction, decQuadFromInt32(&tmp, MICROSECOND_DIGITS), &dec_context);
+            int32_t microsecond = decQuadToInt32Exact(&fraction, &dec_context, DEC_ROUND_DOWN);
+            if (fractional_precision > MICROSECOND_DIGITS) {
+                // Python only supports up to microsecond precision
+                fractional_precision = MICROSECOND_DIGITS;
```
cheqianh, line 878: I didn't change too much things for time stamp API, this is one of the few changes I made.

[Before](https://github.com/cheqianh/ion-python/blob/43bf9243c9b145f8e2baca32624f693bfb3d87ab/amazon/ion/ioncmodule.c#L733) we throw an error if the timestamp value has more than `MICROSECOND_DIGITS` precision. In order to match c spec and not fail tests, now we keep `MICROSECOND_DIGITS` digits and set `DEC_ROUND_DOWN`.

e.g. by `DEC_ROUND_DOWN`, 0.1234567 s will be round to 0.123456 which matches python timestamp round spec.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+            IONCHECK(ion_reader_read_double(hreader, &double_value));
+            py_value = Py_BuildValue("d", double_value);
+            ion_nature_constructor = _ionpyfloat_fromvalue;
+            break;
+        }
+        case tid_DECIMAL_INT:
+        {
+            ION_DECIMAL decimal_value;
+            IONCHECK(ion_reader_read_ion_decimal(hreader, &decimal_value));
+            char dec_str[DEC_NUMBER_MAX_LEN]; // DECQUAD_String is only 43, which fails some tests.
```
cheqianh, line 1050: Default max decimal number is 43. I created a viable DEC_NUMBER_MAX_LEN = 300 to support more decimal digits.

amazon/ion/simpleion.py
```diff
@@ -297,7 +303,7 @@ def dumps(obj, imports=None, binary=True, sequence_as_stream=False, skipkeys=Fal
     return ret_val
 
 
-def load(fp, catalog=None, single_value=True, encoding='utf-8', cls=None, object_hook=None, parse_float=None,
+def load_original(fp, catalog=None, single_value=True, encoding='utf-8', cls=None, object_hook=None, parse_float=None,
```
cheqianh, line 306: I renamed load to load_original so:

1. prevent confusion from the same name API. (pure python load api and the final load api which users use)
2. We may use both pure python load and c extension load in a test.

So does dump.

amazon/ion/simpleion.py
```diff
@@ -453,3 +459,64 @@ def loads(ion_str, catalog=None, single_value=True, encoding='utf-8', cls=None,
...
+                              tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker)
+        except:
+            return dump_original(obj, fp, imports=imports, binary=binary, sequence_as_stream=sequence_as_stream,
+                                 skipkeys=skipkeys, ensure_ascii=ensure_ascii, check_circular=check_circular,
+                                 allow_nan=allow_nan, cls=cls, indent=indent, separators=separators, encoding=encoding,
+                                 default=default, use_decimal=use_decimal, namedtuple_as_object=namedtuple_as_object,
+                                 tuple_as_array=tuple_as_array, bigint_as_string=bigint_as_string, sort_keys=sort_keys,
+                                 item_sort_key=item_sort_key, for_json=for_json, ignore_nan=ignore_nan,
+                                 int_as_string_bitcount=int_as_string_bitcount, iterable_as_array=iterable_as_array,
+                                 tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker, **kw)
```
cheqianh, line 496: Due to beta stage, I used above try...except to handle c extension unexpected issue. If there is some issue for c extension we run the original python API instead.

This try...catch statement will be eventually removed after c extension becomes more stable.

install.py
```diff
@@ -0,0 +1,188 @@
...
+def _build_ionc():
+    if _WIN:
+        _build_ionc_win()
+    elif _MAC or _LINUX:
+        _build_ionc_mac_and_linux()
+
+
+def _build_ionc_win():
+    # check_call('cmake -G \"Visual Studio 15 2017 Win64\"')
+    check_call('cmake -G \"Visual Studio 16 2019\"')
```
cheqianh, line 108: Need to figure out a regular way to compile ionc.

If we are going to leverage GitHub actions to distribute, need to check if the virtual machine GitHub hosts already installed VisualStudio.

setup.py
```diff
@@ -17,31 +17,63 @@
...
+                                  'amazon/ion/ion-c-build/include/ionc',
+                                  'amazon/ion/ion-c-build/include/decNumber'],
+                    libraries=['ionc', 'decNumber'],
+                    library_dirs=['amazon/ion/ion-c-build/lib'],
+                    extra_link_args=['-Wl,-rpath,%s' % '$ORIGIN/ion-c-build/lib',  # LINUX
+                                     '-Wl,-rpath,%s' % '@loader_path/ion-c-build/lib'  # MAC
+                                     ],
+                ),
+            ],
+        )
```
cheqianh, line 43: On linux system,

After building wheels, it's called something like amazon.ion-0.7.95-cp37-cp37m-linux_x86_64.whl while pypi only supports manylinux (right now is linux) wheels. However one solution is just simply rename the wheel to something like amazon.ion-0.7.95-cp37-cp37m-manylinux2010_x86_64.whl...... I'm going to do more research about this.

setup.py
```diff
@@ -17,31 +17,63 @@
...
+    else:
+        print('Using pure python implementation.')
+        kw = dict()
+
+
+    setup(
+        name='amazon.ion',
+        version='0.7.98',
```
cheqianh, line 51: Change to right version: 0.7.0

tests/test_simpleion.py
```diff
@@ -269,14 +270,15 @@ def _assert_symbol_aware_ion_equals(assertion, output):
...
         else:
-            assert (b'$ion_1_0 ' + p.expected) == res
-    else:
-        # The payload contains a LST. The value comes last, so compare the end bytes.
-        assert p.expected == res[len(res) - len(p.expected):]
+            # The payload contains a LST. The value comes last, so compare the end bytes.
+            assert p.expected == res[len(res) - len(p.expected):]
     # test load
     res = loads_func(res, single_value=(not p.stream))
     _assert_symbol_aware_ion_equals(p.obj, res)
```
cheqianh, line 284: Above line 283-284 is how to test load object. Due to different spec between ion-c and ion-python, their dump result (text/binary) are slightly different.
For example: C uses \xd4 to start a struct with length 4 while python uses \xde\x84.

In addition, ion equivalence API used for comparing objects so we can't invoke it here to solve this issue.

There are two ways I think of to solve this issue(there may be more):

We don't care about what result ion-c writes as long as the object that the result loads back is equivalent to the original object.
Adding additional expected values for ion-c in somewhere such as here. If the c extension is on, we compare the dump result to ion-c expected value, otherwise compare it to the original result.
For example:
before: ({a:2}, \xde\x83\x8a21\x02)
now: ({a:2}, \xde\x83\x8a21\x02, \xd3\x8a21\x02)
if c extension is on, we compare the actual result to \xd3\x8a21\x02 otherwise compare it to\xde\x83\x8a21\x02
but by this way, we need to update python unit test whenever ion-c changes.

tests/test_vectors.py
```diff
@@ -117,6 +118,13 @@ def _open(file):
     _equivs_file(u'timestampSuperfluousOffset.10n')  # TODO amzn/ion-python#121
 )
 
+
+if c_ext:
+    _SKIP_LIST += (
+        _good_file(u'subfieldVarInt.ion'),  # c_ext supports 300 decimal digits while there is a 8000+ decimal digits test..
```
cheqianh, line 124: Need to figure out why this file is not in ion-c skip list.

amazon/ion/simpleion.py
```diff
@@ -453,3 +459,64 @@ def loads(ion_str, catalog=None, single_value=True, encoding='utf-8', cls=None,
...
+def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, skipkeys=False, ensure_ascii=True,
+         check_circular=True, allow_nan=True, cls=None, indent=None, separators=None, encoding='utf-8', default=None,
+         use_decimal=True, namedtuple_as_object=True, tuple_as_array=True, bigint_as_string=False, sort_keys=False,
+         item_sort_key=None, for_json=None, ignore_nan=False, int_as_string_bitcount=None, iterable_as_array=False,
+         tuple_as_sexp=False, omit_version_marker=False, **kw):
+    if c_ext and imports is None:
+        try:
+            return dump_extension(obj, fp, binary=binary, sequence_as_stream=sequence_as_stream,
+                              tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker)
+        except:
```
cheqianh, line 488: Due to beta stage, I used above try...except to handle c extension unexpected issue. If there is some issue for c extension we run the original python API instead.

This try...catch statement will be eventually removed after c extension becomes more stable.

cheqianh:
The Travis failed due to wheel build. Will look into this later.

amazon/ion/ioncmodule.c
tgregg: Same question about avoiding string.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+            break;
+        }
+        case tid_SYMBOL_INT:
+        {
+            emit_bare_values = FALSE; // Symbol values must always be emitted as IonNature because of ambiguity with string.
+            ION_STRING string_value;
+            IONCHECK(ion_reader_read_string(hreader, &string_value));
+            ion_nature_constructor = _ionpysymbol_fromvalue;
+            py_value = ion_string_to_py_symboltoken(&string_value);
+            break;
```
tgregg, line 1070: Does this work for symbols with unknown text?

amazon/ion/simpleion.py
tgregg: Is it possible for dump to fail after partially writing to `fp`? If so, this attempt at recovery won't really work.

amazon/ion/simpleion.py
tgregg: Instead of load/dump_original let's use load/dump_pure_python

install.py
```diff
@@ -0,0 +1,188 @@
...
+
+def _library_exists():
+    return _library_exists_helper('ionc') and _library_exists_helper('decNumber')
+
+
+def _library_exists_helper(name):
+    return os.path.exists(join(_C_EXT_DEPENDENCY_LIB_LOCATION, _get_lib_name(name)))
+
+
+def _download_ionc():
```
tgregg, line 69: As discussed offline, let's use an ion-c module so it can be pinned to a particular commit. In this script we'd need to make sure it's checked out, then use the same build logic.

install.py
tgregg: +1

tests/test_cookbook.py
```diff
@@ -234,7 +235,10 @@ def test_write_numeric_with_annotation_simpleion():
     # http://amzn.github.io/ion-docs/guides/cookbook.html#reading-numeric-types
     value = IonPyFloat.from_value(IonType.FLOAT, 123, (u'abc',))
     data = simpleion.dumps(value, binary=False)
-    assert u'$ion_1_0 abc::123.0e0' == data
+    if c_ext:
+        assert u'$ion_1_0 abc::123e+0' == data
+    else:
+        assert u'$ion_1_0 abc::123.0e0' == data
```
tgregg, line 241: Instead of checking whether the C extension is enabled, let's just make it `assert u'$ion_1_0 abc::123.0e0' == data or assert u'$ion_1_0 abc::123e+0' == data`

tests/test_simpleion.py
tgregg: I think we should just compare it against all of the known equivalent representations that we know about, without switching based on whether the C extension is enabled. Just add some more `or` comparisons.

tests/test_vectors.py
tgregg: +1

tgregg (commented):
It's coming along well. I'm excited for the performance improvements.

Once this review is finished, we should read/write some large sample files and try to detect memory leaks using, e.g. Valgrind or whatever the best tool is for memory analysis with Python C extensions. The manual memory management in the C extension is tricky and it's possible we missed something.

tests/test_simpleion.py
cheqianh: **Description:**
Unit test expected value supports tuple now. 

If an expected value is a tuple, the assertion will pass as long as one of the expected values in the tuple are satisfied with assertion conditions.

The timestamp precision for c extension is 6 digits while python supports more, I comment [these tests](https://github.com/cheqianh/ion-python/compare/c_ext_final_review...cheqianh:change#diff-a6ee63a407fe4a0b2f4d9cbf4b74b9e01cd2e72b05eeacc0962bdabf8edfb5bbR214) for now should we support more digits?

**One issue:**
There is one kind of unit test fails. Symbols in the symbol table has different order for ion c and ion python, which causes different binary output. For example:
// annot1::annot2::value
ion_symbol_table::{symbols:["annot1", "annot2", "value"]} $10::$11::$12     //Python
ion_symbol_table::{symbols:["value", "annot1", "annot2"]} $11::$12::$10      //C

The symbol ID is hardcoded in the unit test. For example, in ion python the annotation always come the first so unit test just hardcode something like `\xe5\x82\x8a\x8b \x8c` to represent `$10$11:: $12`, I can add another expected value such as `\xe5\x82\x8b\x8c \x8a` representing `$11$12:: $10` to make it work. but it's not correct if the test case has a symol table with 2 symbols such as:
 ion_symbol_table::{symbols:["annot1", "annot2"]} $10::$11::"a_string".   //C
$12 is out of range.

I manually checked them and they all valid, before I change some lower helper functions to make it work, is there any idea for this? If above description is not clear, feel free to ping me to go through more details verbally.

tests/test_vectors.py
cheqianh: This is because we used string to convert between python decimal and ion decimal, it will be solved once we calculate them without using string.

amazon/ion/simpleion.py
cheqianh: I removed the try..catch branch. 

Instead of some recovery logic, we may just put c extension in the beta stage for a while until it's more stable, is it sounds good?

One of the reasons I tried to put it in the beta stage is because some CPython layer issue such as wrong type conversion will crash python instead of throwing exceptions.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+        );
+    }
+    else {
+        PyList_Append(pyContainer, (PyObject*)element);
+    }
+    Py_XDECREF(element);
+}
+
+/*
+ *  Converts an ion decimal string to a python-decimal-accept string. NOTE: ion spec uses 'd' in a decimal number
```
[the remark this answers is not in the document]
cheqianh: Python decimal accepts string with char 'e'. e.g. `"123e456"`.

`ion_decimal_to_string` API returns string with char 'd'. e.g. `"123d456"` which fails python decimal constructor.

This can be solve after we calculate decimal without using strings.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+
+        if (PyUnicode_Check(key)) {
+            ION_STRING field_name;
+            ion_string_from_py(key, &field_name);
+            IONCHECK(ion_writer_write_field_name(writer, &field_name));
+        }
+        else if (key == Py_None) {
+            // if field_name is None, write symbol $0 instead.
+            IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
+        }
```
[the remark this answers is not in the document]
cheqianh: Above `Py_None` indicates `$0` here. The if...else branch is for 'field_name' with `$0` rather than `null` value. The `null` field_name fails before reaching here. 

More specifically, below are some example:

```.py
obj = "{null: 12}"                      # This will fail, doesn't reach above branch
simpleion.dumps(simpleion.loads(obj), binary=False))    # amazon.ion.exceptions.IonException: IERR_INVALID_FIELDNAME


obj = "{$0: null}"            # This will reach above branch.
simpleion.dumps(simpleion.loads(obj), binary=False))     # {$0:null}
simpleion.dumps(simpleion.loads(obj), binary=True))      # b'\xe0\x01\x00\xea\xd2\x80\x0f'
```

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+static iERR ionc_write_big_int(hWRITER writer, PyObject *obj) {
+    iENTER;
+    PyObject* int_str = PyObject_CallMethod(obj, "__str__", NULL);
+    ION_STRING string_value;
+    ion_string_from_py(int_str, &string_value);
+    ION_INT ion_int_value;
+
+    IONCHECK(ion_int_init(&ion_int_value, NULL));
+    IONCHECK(ion_int_from_string(&ion_int_value, &string_value));
+    IONCHECK(ion_writer_write_ion_int(writer, &ion_int_value));
```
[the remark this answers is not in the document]
cheqianh: I implement converting between arbitrary python int and ion int with math first because converting without string is a little bit tricky. (e.g. need to take care of max int overflow and unsigned ion_int digits, some python math issue such as 18446744073709551615 = (2^62)*3 +1 but 18446744073709551615/(2^62) = 4 and so on..)  

After making sure this change is what we want, I'll implement the rest of "avoiding string" part (decimal, float and so on.).

Write(dumps) code is [here](https://github.com/cheqianh/ion-python/compare/c_ext_final_review...cheqianh:change#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR461)
Read(loads) code is [here](https://github.com/cheqianh/ion-python/compare/c_ext_final_review...cheqianh:change#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR1083)

amazon/ion/ioncmodule.c
cheqianh: As discussed offline, we will use pure python implementation to handle `catalog` and `import` options and support more options incrementally.

cheqianh:
#### 1. What I did:
- [x] **Avoid using string to improve performance.**  Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618688221) for details.
- [x] **Changed Unit test that able to test both ion-python and ion-c spec.** Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618651040) for details.
- [x] **Removed try..catch branch for simple-ion dump/load.** Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618653061) for details.
- [x] **One case failed python test but not in C skip-list.** Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618651164) for details.
- [x] **Other comments.**
        1. Py decimal accept string. Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618678939)
        2. Null field name. Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618687018)
        3. Unknown text symbol. Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618700772)


#### 2. Next step: 
Automate process of building wheels. Some configuration issues will be solved there. I am going to create a new Pull Request targeting this one and implement for below platforms.
- [ ] Windows 
- [ ] Mac 
- [ ] Linux Family 

#### 3. After finishing automation part:
- [ ] Run large files and do a memory leak check. (important)
- [ ] Some TODOs in the code.
- [ ] Fix an [ion-c issue](https://github.com/amzn/ion-c/issues/223) and more issues as time allows. 


Let me know if there is anything need to be prioritized!

tests/test_simpleion.py
tgregg: Should we just change the tests to compare the text and not symbol ID?

amazon/ion/simpleion.py
tgregg: Yeah, let's just rely on beta testing.

amazon/ion/ioncmodule.c
```diff
@@ -458,16 +460,80 @@ static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_
...
+
+    ION_INT ion_int_value;
     IONCHECK(ion_int_init(&ion_int_value, NULL));
-    IONCHECK(ion_int_from_string(&ion_int_value, &string_value));
+
+    // Determine sign
+    if (PyObject_RichCompareBool(temp, PyLong_FromLong(0), Py_LT) == 1) {
+        ion_int_value._signum = -1;
+        temp = PyNumber_Negative(temp);
+    } else if (PyObject_RichCompareBool(temp, PyLong_FromLong(0), Py_GT) == 1) {
```
tgregg, line 475: Should `PyLong_FromLong(0)` be stored in a variable to avoid recomputing it? Or is 0 a special case that requires no computation?

amazon/ion/ioncmodule.c
```diff
@@ -458,16 +460,80 @@ static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_
...
+
+        if (pow_value == Py_None) {
+            // pow(2^31, base) should be calculated correctly.
+            _FAILWITHMSG(IERR_INTERNAL_ERROR, "Calculation failure: 2^31 .");
+        }
+
+        // Python equivalence: digit = temp / 2^31, temp = temp % 2^31
+        res = PyNumber_Divmod(temp, pow_value);
+        py_digit = PyNumber_Long(PyTuple_GetItem(res, 0));
+        py_reminder = PyTuple_GetItem(res, 1);
```
tgregg, line 505: `py_reminder` -> `py_remainder`

amazon/ion/ioncmodule.c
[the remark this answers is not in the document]
tgregg: Looks good to me.

amazon/ion/ioncmodule.c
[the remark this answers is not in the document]
tgregg: Got it, thanks.

tests/writer_util.py
```diff
@@ -76,26 +81,26 @@ def _convert_clob_pairs(clob_pairs):
...
     (u'_hello_world', br"_hello_world"),
     (u'null', br"'null'"),
     (u'hello world', br"'hello world'"),
     (u'hello\u0009\x0a\x0dworld', br"'hello\t\n\rworld'"),
-    (u'hello\aworld', br"'hello\x07world'"),
-    (u'hello\u3000world', br"'hello\u3000world'"), # A full width space.
-    (u'hello\U0001f4a9world', br"'hello\U0001f4a9world'"), # A 'pile of poo' emoji code point.
+    (u'hello\aworld', (br"'hello\x07world'", br"'hello\aworld'")),
+    (u'hello\u3000world', (br"'hello\u3000world'", b"'hello\xe3\x80\x80world'")),  # A full width space.
+    (u'hello\U0001f4a9world', (br"'hello\U0001f4a9world'", b"'hello\xf0\x9f\x92\xa9world'")),  # A 'pile of poo' emoji code point.
```
tgregg, line 93: I don't think the `\x` representations are equivalent in this case. `\u` and `\U` indicate  unicode code points, while `\x` is just for raw bytes. Maybe there's a flaw in ion-c's unicode escaping logic?

cheqianh: As discussed offline,  I noticed that 'b' prefix treats bytes as a whole here.

I did some experiment that shows unicode representation of `u'\U0001f4a9'` is equivalent to `b'\xf0\x9f\x92\xa9'`
 
```.py
a = u'\U0001f4a9'.encode('utf-8')
b = b'\xf0\x9f\x92\xa9'
print(a)    # 💩
print(b)      # 💩
print(a == b)    # True
```

amazon/ion/ioncmodule.c
cheqianh: Changed.

amazon/ion/ioncmodule.c
cheqianh: Changed.

tests/test_vectors.py
cheqianh: Our current algorithm of ion/python decimal conversion only support 34 digits. So we may still want to put it in skip list. 

Looked into ion-c tests, this is a good vector and ion-c read value as an event_stream. So the value is stored as a string in an ion_event and doesn't raise an error.

tests/test_simpleion.py
cheqianh: The test failed for binary type and expected values are hardcoded since ion-python annotations are always the smallest SID above 10 in the symbol table. (e.g. `annot1` and `annot2` are already $10 and $11)

The expected values also start at annotation part and followed by values (not include symbol table).

An alternate solution is changing ion-c symbol table order to write annotations first, but if this is not a simple change then it may not be a good solution since the unit tests just failed due to "string comparison" rather than data module equivalence.

amazon/ion/ioncmodule.c
cheqianh: As discussed offline, we only support 34 decimal digits. There is a test with more than 34 digits throws NUMBER_FLOW exception. 

```
_IT.DECIMAL: (
(_D('1.1999999999999999555910790149937383830547332763671875e0'),   # <- value
 b'1.1999999999999999555910790149937383830547332763671875'))           # <- expected value
```

dump/write decimal function [here](https://github.com/amzn/ion-python/pull/149/commits/a176f6f483191aa6c9907dc4e66af22bbd912af9#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR633).
load/read decimal function [here](https://github.com/amzn/ion-python/pull/149/commits/a176f6f483191aa6c9907dc4e66af22bbd912af9#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR1181).

cheqianh:
#### Summary:
1. **Avoid using string for ion/python decimal conversion**: Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618651040)
2. **Unit test for different order symbol IDs**: Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618651040)
3. **One vector not in ion-c skip list fails c extension unit test**: Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r618651164)
4. **Unicode and raw bytes representation in unit test**: Refer to [here](https://github.com/amzn/ion-python/pull/149#discussion_r633868100)
5. **Other changes.** such as typo remainder, re-use PyLong_FromLong(0) etc.

install.py
cheqianh: Will work on this in the "releasing beta stage" work.

setup.py
cheqianh: Will work on this in the "releasing beta stage" work

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+ *  Builds a python string using an ION_STRING
+ *
+ *  Args:
+ *      string_value:  An ION_STRING that needs to be converted
+ *
+ *  Returns:
+ *      A python string
+ */
+static PyObject* ion_build_py_string(ION_STRING* string_value) {
+    // TODO Test non-ASCII compatibility.
```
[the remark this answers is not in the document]
cheqianh: Will work on TODOs in the future step.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+    buf = (BYTE*)(PyMem_Malloc((size_t)len));
+    SIZE bytes_read;
+    IONCHECK(ion_stream_read(ion_stream, buf, (SIZE)len, &bytes_read));
+
+    IONCHECK(ion_stream_close(ion_stream));
+    if (bytes_read != (SIZE)len) {
+        FAILWITH(IERR_EOF);
+    }
+    // TODO Py_BuildValue copies all bytes... Can a memoryview over the original bytes be returned, avoiding the copy?
+    PyObject* written = Py_BuildValue(IONC_BYTES_FORMAT, (char*)buf, bytes_read);
```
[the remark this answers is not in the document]
cheqianh: Will work on TODOs in the future step.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+        IONCHECK(err);
+    }
+    else {
+        IONCHECK(ionc_write_value(writer, obj, tuple_as_sexp));
+    }
+    IONCHECK(ion_writer_close(writer));
+
+    POSITION len = ion_stream_get_position(ion_stream);
+    IONCHECK(ion_stream_seek(ion_stream, 0));
+    // TODO if len > max int32, need to return more than one page...
```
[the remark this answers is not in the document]
cheqianh: Will work on TODOs in the future step.

amazon/ion/ioncmodule.c
```diff
@@ -0,0 +1,1369 @@
...
+    }
+    int ion_type = ION_TYPE_INT(t);
+
+    switch (ion_type) {
+        case tid_EOF_INT:
+            SUCCEED();
+        case tid_NULL_INT:
+        {
+            ION_TYPE null_type;
+            // Hack for ion-c issue https://github.com/amzn/ion-c/issues/223
```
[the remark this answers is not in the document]
cheqianh: Will work on TODOs in the future step.

install.py
cheqianh: Will work on this in next step.

tgregg (approved):
Looks good. Do you know why the Travis builds are failing?

cheqianh:
highlights:
1. Travis failed because it builds c extension without initialize ion-c binaries. 
2. Now all the setup work can be done in `setup.py` and no need to run `install.py` separately. 
3. Travis is still failing due to two kinds of tests: decimal with more than 34 digits and different order SID.

cheqianh:
Next step:
- [ ] Go through all the tests again and do some minor change.
- [ ] Create a new branch `c-extension-beta-stage` and change the current PR targeting new branch.
- [ ] Work on install.py.

cheqianh:
The travis fails due to known unit test failures and manually checked they are all equivalence. I'll push it to c-extension-beta first. 

I will create a new Pull Request to finish all the left action items.

Issue #160, opened 2021-07-07
C extension only supports 6 timestamp precision.


cheqianh:
Right now C extension only supports at most 6 for timestamp precision, it would be nice to support more.

cheqianh:
It would be good to support at least **9** digits for QLDB users.

cheqianh:
Above commit changed C extension to support 9 digits subseconds now.

PR #164, opened 2021-08-27
C extension supports 9 timestamp precision.

cheqianh:
## Description:
Solved #160 --- C extension supports 9 timestamp precision now. Since Python's datetime only supports microsecond precision so this change only affects ion [timestamp](https://github.com/amzn/ion-python/blob/d3ab7c0d12a76df9356b296981276480ec6a9f59/amazon/ion/core.py#L394).

**Code details:**
When reading a timestamp, if it has more than 6 precision, constructing an ion timestamp by passing [fractional_seconds](https://github.com/amzn/ion-python/blob/d3ab7c0d12a76df9356b296981276480ec6a9f59/amazon/ion/core.py#L411) argument and otherwise using microseconds as before.

When writing a timestamp to a file. Reading at most 9 decimal digits that from python ion timestamp into C decQuad and using ion-c API to dump them.

## Test:
✅ &nbsp;**Memory leak test**: No leak.
✅ &nbsp;**Performance measurement**: timeit() shows that It's almost the same speed as before.
✅ &nbsp;**Unit test**:  Screen Shot 2021-08-26 at 12 09 47 PM

&nbsp;
&nbsp;

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -726,10 +726,10 @@ iERR ionc_write_value(hWRITER writer, PyObject* obj, PyObject* tuple_as_sexp) {
         }
 
         ION_TIMESTAMP timestamp_value;
-        PyObject* fractional_seconds, *fractional_decimal_tuple, *py_exponent;
+        PyObject *fractional_seconds, *fractional_decimal_tuple, *py_exponent, *py_digits;
         int year, month, day, hour, minute, second;
         short precision, fractional_precision;
-        short final_fractional_precision, final_fractional_seconds;
+        int final_fractional_precision, final_fractional_seconds;
         if (PyObject_HasAttrString(obj, "precision")) {
             // This is a Timestamp.
             precision = int_attr_by_name(obj, "precision");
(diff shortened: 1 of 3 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -994,31 +1038,50 @@ static iERR ionc_read_timestamp(hREADER hreader, PyObject** timestamp_out) {
...
+                char decimal[fractional_precision];
+                sprintf(decimal, "%d", dec);
+                char* dec_num = string_concatenation(2, fractional_precision, front, decimal);
+
+                PyObject* py_dec_str = PyUnicode_FromString(dec_num);
+                PyObject* py_fractional_seconds =
+                    PyObject_CallFunctionObjArgs(_decimal_constructor, py_dec_str, NULL);
+                PyDict_SetItemString(timestamp_args, "fractional_seconds", py_fractional_seconds);
+                Py_DECREF(py_fractional_seconds);
+                Py_DECREF(py_dec_str);
```
tgregg, line 1068: Rather than creating a temporary string to be re-parsed by the decimal constructor, we should just create a decimal from the `dec` integer, then `scaleb` it using `fractional_precision`.

tgregg (commented):
Looks pretty good. One comment below.

amazon/ion/ioncmodule.c
cheqianh: Changed. Passed unit test.

amazon/ion/ioncmodule.c
```diff
@@ -1055,14 +1047,14 @@ static iERR ionc_read_timestamp(hREADER hreader, PyObject** timestamp_out) {
                     decContextClearStatus(&dec_context, DEC_Inexact);
                 }
 
-                char* front = "0.";
-                char decimal[fractional_precision];
-                sprintf(decimal, "%d", dec);
-                char* dec_num = string_concatenation(2, fractional_precision, front, decimal);
+                char* dec_num = malloc(fractional_precision + 2);
```
tgregg, line 1050: Will this be freed when `py_dec_str` below is garbage collected, or does this need to be freed manually after `PyObject* py_dec_str = PyUnicode_FromString(dec_num);` to avoid a memory leak?

cheqianh: Ah Thanks.The memory leak test didn't detect any issue and I did some investigation on it. C will eventually de-allocate the memory we allocated after completion of the execution of the program if we don't call free(). However, It's still good to have a free() here in favor of applications that run very long time so we can always free them right after we use them.

tgregg (approved):
Squash and merge.

PR #172, opened 2021-09-22
Update simpleion.load function to support iterating infinite streams

alimg:
*Issue #, if available:* N/A

*Description of changes:* Converted the ionc_read implementation to return and iterator instance to allow processing large ION streams without having to load entire input buffers into memory. 

Sharing this pull request to see if there's interest to see these changes merged. If so, I can work on few improvements to this changeset to add more integration tests and verify there are no memory leak with valgrind etc..

*Testing:* Tested with unit test and made sure they all pass. And used this version briefly in a separate workload to process some big files.

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -1,3 +1,6 @@
+// See https://python.readthedocs.io/en/stable/c-api/arg.html#strings-and-buffers
+#define PY_SSIZE_T_CLEAN
+
 #include <Python.h>
 #include "datetime.h"
 #include "_ioncmodule.h"
@@ -18,6 +21,8 @@
 #define FIELD_NAME_MAX_LEN 1000
 #define ANNOTATION_MAX_LEN 50
 
+#define IONC_STREAM_READ_BUFFER_SIZE 1024
+
 static char _err_msg[ERR_MSG_MAX_LEN];
 
 #define _FAILWITHMSG(x, msg) { err = x; snprintf(_err_msg, ERR_MSG_MAX_LEN, msg); goto fail; }
@@ -25,7 +30,7 @@ static char _err_msg[ERR_MSG_MAX_LEN];
 // Python 2/3 compatibility
 #if PY_MAJOR_VERSION >= 3
     #define IONC_BYTES_FORMAT "y#"
-    #define IONC_READ_ARGS_FORMAT "OOO"
+    #define IONC_READ_ARGS_FORMAT "OO"
     #define PyInt_AsSsize_t PyLong_AsSsize_t
     #define PyInt_AsLong PyLong_AsLong
     #define PyInt_FromLong PyLong_FromLong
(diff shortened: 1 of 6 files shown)


tgregg:
@alimg This looks like a great idea to me; yes, we have interest in seeing these changes merged. We will work on scheduling time to review this PR in detail.

cheqianh (commented):
This is a really good change, Thanks for the contribution!

&nbsp;
For your comment:
> Sharing this pull request to see if there's interest to see these changes merged. If so, I can work on few improvements to this changeset to add more integration tests and verify there are no memory leak with valgrind etc..

Yes we are interested to see this PR got merged. Could you add some integration tests and essential unit tests to make sure the iterator functions correctly? 

In addition, I did some pre-profiling 
1. I did some performance measurement and it's almost as fast as the previous C extension, which is good.
2.  I did some memory leak profiling (I know you might do profiling too, I just write it down here to keep this on track). It seems that the `ionc_read` method has some memory leak issue when `parse_eagerly` equals to both true and false. So I think the leak might happen somewhere in the iterator's setup/next methods other than ion_c parse level. Could you please take a look at this?

alimg:
Updated the PR to fix a leak in dealloc method and some some other leaks that were revealed mostly around exceptional cases. I used `PYTHONMALLOC=malloc valgrind --show-leak-kinds=definite --leak-check=full  ` --addopts "-s tests/test_simpleion.py "` command-line to check for memory leaks. Full test suite looks mostly clean now but there are still some illegal memory accesses and small sources of leaks. I'll try to take a stab at them another time.

amazon/ion/ioncmodule.c
```diff
@@ -1047,7 +1103,7 @@ static iERR ionc_read_timestamp(hREADER hreader, PyObject** timestamp_out) {
                     decContextClearStatus(&dec_context, DEC_Inexact);
                 }
 
-                char* dec_num = malloc(fractional_precision + 2);
+                char* dec_num = malloc(fractional_precision + 6); // TODO: decQuadToString call below seems to overflow the allocated memory
```
alimg, line 1106: Valgrind revealed some illegal memory access made by the 3rd party code where it tried to write beyond the number of bytes we allocated here. I've bumped up the padding a bit, but still not sure if this is the right way

cheqianh (commented):
Thanks for the contribution! I only had some minor questions and left comments below.

I left some comments about documentation and tests, please feel free to change them if you would like to. Otherwise I would also be happy to create a pull request later to add little more tests and description.

For two [memory leaks](https://github.com/alimg/ion-c/commit/f8add5ed1bfee0deec63926af94bad65fd6f6d29) found in ion-c, I saw a similar issue opened earlier and it's fixed in [this PR](https://github.com/amzn/ion-c/pull/257), I'll take a look at the other one.

I added a [commit](https://github.com/amzn/ion-python/pull/172/commits/1c3e0002f710b4b0a7cfcd14e5b6cf97f0e95cc7) to fix the conflict due to the recent change in c-extension-beta branch.

cheqianh (approved):

amazon/ion/ioncmodule.c
alimg: Dumped some info from my environment [here](https://gist.github.com/alimg/632440698bcf8949f8b6fc8552de9670) but I must admit I don't know what could be platform dependent here.

Is there a reason not to use [DECQUAD_String](https://github.com/amzn/ion-c/blob/a541a810b8e0471262f7bfda04283859a07b7d10/decNumber/example7.c#L18) here? I've updated this to match the other examples used in the ion-c code base, but I hope I didn't remove any specific memory saving trick put in place before.

amazon/ion/ioncmodule.c
```diff
@@ -1474,46 +1532,158 @@ iERR ionc_read_all(hREADER hreader, PyObject* container, BOOL in_struct, BOOL em
...
+
+fail:
+    Py_XDECREF(container);
+    PyObject* exception = PyErr_Format(_ion_exception_cls, "%s %s", ion_error_to_str(err), _err_msg);
+    _err_msg[0] = '\0';
+    return exception;
+}
+
+PyObject* ionc_read_iter(PyObject *self) {
+    ionc_read_Iterator *iterator = (ionc_read_Iterator*) self;
```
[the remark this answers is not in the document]
alimg: no reason at all, I meant to delete this line

tests/test_simpleion.py
```diff
@@ -612,6 +612,24 @@ def test_roundtrip(p):
...
+@parametrize(
+    *tuple(_generate_roundtrips(_ROUNDTRIPS))
+)
+def test_roundtrip_ion_stream(p):
+    obj, is_binary, indent, tuple_as_sexp = p
+    expected  = [obj]
+    out = BytesIO()
+    dump(obj, out, binary=is_binary, indent=indent, tuple_as_sexp=tuple_as_sexp)
+    out.seek(0)
+    res = load(out, single_value=False, parse_eagerly=True)
```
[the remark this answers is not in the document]
alimg: Yes, I added this to make sure the new code paths on the Python side are tested. 

When `parse_eagerly is True`, lazy mode always returns an iterator, even if there was only one ion value in the input stream. I wanted to capture this in the tests.

alimg:
>     3. I added a [commit](https://github.com/amzn/ion-python/pull/172/commits/1c3e0002f710b4b0a7cfcd14e5b6cf97f0e95cc7) to fix the conflict due to recent change in the c-extension-beta branch.
Rebased from the upstream to address this.

cheqianh:
Thanks for the rebase. You might not have to rebase it everytime because we can use `Squash and merge` button below to squash all commits into one later.

amazon/ion/ioncmodule.c
cheqianh: Yeah I agree, using `DECQUAD_String` is a good idea. Change `+2` to `+6` is not a big deal too. I'm just curious dump what data causes the illegal memory access.

Issue #186, opened 2022-01-20
C extension distribution enhancement 


cheqianh:
It would be good to do more investigation on below items:

1. Now we distribute c binaries to PYPI too but the script will clean them up during installation. It would be good to find a way to **NOT** distribute binaries to PYPI but build these binaries after installation and includes them in the final directory.

2. Using git clone instead of git submodule, figure out why submodule doesn't work during installation.

3. We use string conversion for Ion int/decimal for now since Windows require a VS C++ build tool which otherwise raise errors. 

4. jsonconversion is added to` requirement.txt` but not in [setup.py's install_requires](https://github.com/amzn/ion-python/blob/master/setup.py#L67). Otherwise it raises an error on windows: `Paclage jsonconversion requires a different Python 3.7.8 not in <= 2.7`.

PR #187, opened 2022-01-20
Fixes distribution issues so the C extension could be enabled by default.

cheqianh:
In order to enable C extension by default, this PR did below changes:

1. Changed back to string-number conversion (I comment the current math logic for reference and use the original string conversion way instead). Windows require a VS C++ build tool which raise some allocation errors. Using string conversion works well. 
2. Changed `git submodule` command to `git clone`.
3. Removed jsonconversion from [setup.py's install_requires](https://github.com/amzn/ion-python/blob/master/setup.py#L67), which is added into requirement.txt, This fails installation on windows with an error msg:`Package jsonconversion requires a different Python 3.7.8 not in <= 2.7`

I opened [an issue](https://github.com/amzn/ion-python/issues/186) for above three changes, not urgent but would be good to do more investigation in the future.


setup.py
```diff
@@ -64,7 +64,6 @@ def run_setup():
 
         install_requires=[
             'six',
-            'jsonconversion'
```
tgregg, line 67: What are the implications of this? Does it break something else? Why was it added in the first place?

tgregg (commented):
Can you go into more detail about the errors you're seeing when you use the number conversion logic that doesn't require intermediate strings? I'd feel much more comfortable temporarily disabling this for Windows only, if necessary, than removing it entirely. Even better would be to determine the source of the errors and swap logic based on OS in as granular a way as possible.

setup.py
cheqianh: Discussed offline, I'll create a new commit and comment detail there.

cheqianh:
1. Removed install_requires, everything works well.
2. Switch to string conversion, commit ed54575 includes conversion logic. I also updated [issue 186](https://github.com/amzn/ion-python/issues/186)

tgregg (approved):

PR #199, opened 2022-05-05
Adds Github Actions for release automation.

cheqianh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,182 @@
+name: Ion Python Release
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+      - "v*.*.*"
+
+jobs:
+  test:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ '3.6', '3.7', '3.8', '3.9', '3.10', 'pypy-3.6', 'pypy-3.7', 'pypy-3.8' ]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up python
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Create a virtual environment
+        run: git submodule init && git submodule update && python3 -m venv ./venv && . venv/bin/activate
+      - run: pip install -r requirements.txt
+      - run: pip install -e .
+      - run: py.test
+
+  source-distribution:
+    if: startsWith(github.ref, 'refs/tags/')
+    name: Source distribution
+    runs-on: ubuntu-latest
+    needs: [test]
+    strategy:
+      matrix:
+        python-version: [ '3.10' ]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Build sdist
+        run: python3 setup.py sdist
+
+      - name: Upload source to Github
+        uses: actions/upload-artifact@v2
+        with:
+          name: source
+          path: ./dist/amazon.ion-*.tar.gz
+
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,182 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Create a virtual environment
+        run: git submodule init && git submodule update && python3 -m venv ./venv && . venv/bin/activate
+      - run: pip install -r requirements.txt
+      - run: pip install -e .
+      - run: py.test
+
+  source-distribution:
```
cheqianh, line 29: Note that we automated the original source release too

.github/workflows/release.yml
```diff
@@ -0,0 +1,182 @@
...
+
+            v=$( pip show amazon.ion | grep Version )
+            break
+          done
+
+          if [[ $v != "Version: ${GITHUB_REF##*/v}" ]]
+          then
+            echo "Exiting because unable to install the new version from PYPI"
+            exit 1
+          fi
```
cheqianh, line 86: This checks if we can install source from PYPI after 5 mins. Note that `--no-binary :all:` is used for installing source only.

.github/workflows/release.yml
```diff
@@ -0,0 +1,182 @@
...
+    strategy:
+      matrix:
+        python-version: [ '3.10' ]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Build sdist
+        run: python3 setup.py sdist
+
+      - name: Upload source to Github
```
popematt, line 43: Where is this uploading to? A release?

.github/workflows/release.yml
```diff
@@ -0,0 +1,182 @@
+name: Ion Python Release
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+      - "v*.*.*"
```
popematt, line 7: I know `ion-dotnet` uses this event for the workflow, but have you considered using the [`release`](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release) event?

tgregg (approved):
Looks good to me, pending resolution of Matthew's comments.

.github/workflows/release.yml
cheqianh: Looked into the document, seems like **release event** covers **tag** situation and it's more focusing on release. I think it's a good idea to use release, I can also send a PR to change ion-dotnet later.

.github/workflows/release.yml
cheqianh: We upload source to GitHub Actions so we can install the built source/wheels directly from Github. E.g, we upload test-driver output to [ion-java GHA](https://github.com/amzn/ion-java/actions/runs/2099291871) so users can download the output when workflow fails, you will find them in the artifacts section at the bottom of the page.

The reason we upload them to Github is because later on we will [download](https://github.com/amzn/ion-python/pull/199/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R141-R143) them from GitHub and zip all wheels in a separate job and upload them to AWS. Since we built wheels on different instances with different OS so I tried to avoid different upload behaviors (E.g. `zip` command works on Mac but Windows use some other command). Therefore I try to gather all wheels in one instance and release them in case some wheels are lost during the release process.

.github/workflows/release.yml
cheqianh: Addressed in below commit.

popematt (approved):

PR #206, opened 2022-07-11
Initializes variables outside of the for loop.

cheqianh:
**Description:**
Some C compilers complain that initializing variables within a for loop is invalid, move it outside.

**Test:**
see GHA below.


nirosys:
It is a pretty good practice to use loop declarations when appropriate. C has supported declarations in the init portion of a for loop since C99, and most compilers in the past 10+ yrs have had support for C99. Though older compilers that do not default to C99 may require a flag like `-std=c99` (clang & gcc), in order to force the compiler to use C99.

Is there a need to support older compilers that do not have C99 support?

cheqianh:
Nice, adding compile args for python setup scripts works 😄, which makes the code clean.

tgregg (approved):

nirosys (approved):

Issue #213, opened 2022-08-06
Python C extension throws IERR_READ_ERROR sometimes for nested struct.


cheqianh:
When reading large data (not necessarily to be large but file size matters), C extension runs into IERR_READ_ERROR.

The root is because of [IONC_STREAM_READ_BUFFER_SIZE](https://github.com/amzn/ion-python/blob/master/amazon/ion/ioncmodule.c#L24) and [safe-guarding check](https://github.com/amzn/ion-python/blob/master/amazon/ion/ioncmodule.c#L1429-L1432). Since iterator is not quiet stable today, I changed some logic to avoid relying on iterator to return Ion objects, and emphasizing in the document that this feature is considered experimental. See PR - https://github.com/amzn/ion-python/pull/214

To reproduce the issue, try using simpleIon to loads data below:
```
"""[
{
  test_area:en_US,
  value:'''test test test test test test test test \u2013 test test test test test test test''',
  value_2:{
    value_3:testest,
    value_4:996
  }
},
{
  test_area:en_US,
  value:'''test test test test test test test test \u2013 test test test test test test test''',
  value_2:{
    value_3:testest,
    value_4:996
  }
},
{
  test_area:en_US,
  value:'''test test test test test test test test \u2013 test test test test test test test''',
  value_2:{
    value_3:testest,
    value_4:996
  }
},
{
  test_area:en_US,
  value:'''test test test test test test test test \u2013 test test test test test test test''',
  value_2:{
    value_3:testest,
    value_4:996
  }
},
{
  test_area:en_US,
  value:'''test test test test test test test test \u2013 test test test test test test test''',
  value_2:{
    value_3:testest,
    value_4:996
  }
},
{
  test_area:en_US,
  value:'''test test test test test test test test \u2013 test test test test test test test''',
  $ims_state:{
    value_3:testest,
    changed_at_version:52
  }
},
]"""
```
It throws `amazon.ion.exceptions.IonException: IERR_READ_ERROR`.

cheqianh:
Similar issue - https://github.com/amzn/ion-python/issues/205

Issue #215, opened 2022-08-18
C extension should allows user to config read bytes size for performance reason 


cheqianh:
Refer to 

> The performance depends on the size of files we are going to read. On the one hand, if the file is really small but we maintain a large buffer, it’s not efficient and slower in performance. On the other hand, if the file is really large but we only read a small number of bytes, say 32 bytes, each time it’s also very slow. I use 32*1024 first and we should eventually allow user to configure this in the future.

If we can control the size of read buffer, users will see around 5-10 % performance improvement (A brief benchmark data for sample file I tested).

Issue #205, opened 2022-06-27
C extension iterator type doesn't read special char correctly within a multiple keys' container 


cheqianh:
To reproduce this, 
```
with open("file_below") as file:
    content = ion.load(file)
    print(content)
```
read file below,
```.py
{
  field1:[
    {
      language_tag:en_US,
      value:"💗This fails to parse only when the string is long enough? I don't fully understand what is going on. The field2 are impacting the parsing of this string?"
    },
    {
      language_tag:en_US,
      value:"Here is the weird apostrophe -> ’   This file can be parsed if the special characters are removed or if some of these strings are shortened or removed..."
    }
  ],
  field2:[
    {
      language_tag:en_US,
      value:"Why does this information interfere with the field1 - I have to put words here to demonstrate the error - testtesttesttesttesttesttesttesttesttesttesttesttesttest"
    },
    {
      language_tag:en_US,
      value:"Why does this information interfere with the field1 - I have to put words here to demonstrate the error - testtesttesttesttesttesttesttesttesttesttesttesttesttest"
    },
    {
      language_tag:en_US,
      value:"Why does this information interfere with the field1 - I have to put words here to demonstrate the error - testtesttesttesttesttesttesttesttesttesttesttesttesttest"
    }
  ]
}
```
it will raise something like,
```
IERR_READ_ERROR
```

cheqianh:
Related to https://github.com/amzn/ion-python/issues/213

PR #216, opened 2022-08-18
Adjests read buffer size to avoid unicode/utf-8 conversion issue.

cheqianh:
### Issue 
This PR fixes https://github.com/amzn/ion-python/issues/213 and https://github.com/amzn/ion-python/issues/205

### Description: 
The root is because of unicode and UTF-8 conversion. 
For example, when we read a String IO including value `\u2013` by `fp.read(1)`, It will be converted to 3 bytes instead 1 later.
which failed [memory bounds check](https://github.com/amzn/ion-python/blob/master/amazon/ion/ioncmodule.c#L1429-L1432). 

Unicode detail, see https://www.fileformat.info/info/unicode/char/2013/index.htm

### Solution
To avoid this, we maintain a 1024x32 bytes buffer, and only read 1024*32/4 bytes each time (since one code point can be up to 4 UTF-8 bytes)

### Benchmark
The performance depends on the size of files we are going to read. On the one hand, if the file is really small but we maintain a large buffer, it’s not efficient and slower in performance. On the other hand, if the file is really large but we only read a small number of bytes, say 32 bytes, each time it’s also very slow. I use 32*1024 first and we should eventually allow user to configure read buffer size for their user case in the future.

I opened an issue to keep this on track - https://github.com/amzn/ion-python/issues/215

### Test
Passed unit test, and run specific test cases that cause the issue. 

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -1427,7 +1427,7 @@ iERR ion_read_file_stream_handler(struct _ion_user_stream *pstream) {
 
     // safe-guarding the size variable to protect memcpy bounds
     if (size < 0  || size > IONC_STREAM_READ_BUFFER_SIZE) {
-        _FAILWITHMSG(IERR_READ_ERROR, "illegal bytes number has been read");
+        FAILWITH(IERR_READ_ERROR);
     }
     memcpy(stream_handle->buffer, char_buffer, size);
 
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -21,7 +21,7 @@
 #define FIELD_NAME_MAX_LEN 1000
 #define ANNOTATION_MAX_LEN 50
 
-#define IONC_STREAM_READ_BUFFER_SIZE 1024
+#define IONC_STREAM_READ_BUFFER_SIZE 1024*32
 
 static char _err_msg[ERR_MSG_MAX_LEN];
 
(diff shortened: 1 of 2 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -1669,7 +1669,8 @@ PyObject* ionc_init_module(void) {
 
     decContextDefault(&dec_context, DEC_INIT_DECQUAD);  //The writer already had one of these, but it's private.
 
-    _arg_read_size = PyLong_FromLong(IONC_STREAM_READ_BUFFER_SIZE);
+    // Divide by four to avoid size overflow due to Unicode and UTF-8 conversion
+    _arg_read_size = PyLong_FromLong(IONC_STREAM_READ_BUFFER_SIZE/4);
```
tgregg, line 1673: Consider storing this size in a constant as well.

tgregg (commented):
Please commit a unit test that exercises the failure before this fix. Something like the example in https://github.com/amzn/ion-python/issues/205

cheqianh:
Refactor the constant and Added a unit test, I also improve the error message so user will got clue why it fails now.

amazon/ion/ioncmodule.c
```diff
@@ -1428,7 +1427,7 @@ iERR ion_read_file_stream_handler(struct _ion_user_stream *pstream) {
 
     // safe-guarding the size variable to protect memcpy bounds
     if (size < 0  || size > IONC_STREAM_READ_BUFFER_SIZE) {
-        FAILWITH(IERR_READ_ERROR);
+        _FAILWITHMSG(IERR_READ_ERROR, "illegal bytes number has been read");
```
tgregg, line 1430: Is this effectively an illegal state? If so, we don't really need the message. Could a user ever encounter this except in the case where there is a bug in the ion-python C extension?

(If we keep the message: "illegal bytes number has been read" -> "An illegal number of bytes have been read.")

tgregg (approved):

PR #220, opened 2022-12-03
Remove support for Python 2.x

rmarrowstone:
- Remove from __future__ imports
- remove six wip
- six removal wip ii
- complete six removal

*Issue #, if available:*

*Description of changes:*


amazon/ion/reader.py
```diff
@@ -122,12 +88,14 @@ def __init__(self, is_unicode=False):
...
+            self.__element_type = str
         else:
-            self.__chr = chr if six.PY2 else lambda x: x
-            self.__element_type = six.binary_type
-        self.__ord = ord if (six.PY3 and is_unicode) else lambda x: x
+            # TODO: this may be able to be fully eliminated
+            self.__chr = lambda x: x
+            self.__element_type = bytes
+        # TODO: this may be able to be fully eliminated
+        self.__ord = ord if is_unicode else lambda x: x
```
tgregg, line 98: Not sure we can eliminate these. My recollection is that we use the BufferQueue for both binary and text data. When it's binary, we need to pass raw bytes back and forth without interpretation. Did you have something else in mind?

amazon/ion/writer_text.py
```diff
@@ -448,7 +441,8 @@ def raw_writer(indent=None):
     if not (indent is None or is_whitespace_str):
         raise ValueError('The indent parameter must either be None or a string containing only whitespace')
 
-    indent_bytes = six.b(indent) if isinstance(indent, str) else indent
+    # latin-1 is what six.b() was encoding this as before
+    indent_bytes = indent.encode("latin-1") if isinstance(indent, str) else indent
```
tgregg, line 445: Let's try UTF-8

amazon/ion/writer_text.py
```diff
@@ -81,13 +73,14 @@ def _serialize_scalar_from_string_representation_factory(type_name, types, str_f
     def serialize(ion_event):
         value = ion_event.value
         validate_scalar_value(value, types)
-        return six.b(str_func(value))
+        # TODO latin-1 is what six.b was doing. it does not seem right to me either
+        return str_func(value).encode("latin-1")
```
tgregg, line 77: I agree... what happens if we change to UTF-8?

tgregg (approved):
What a day!

amazon/ion/writer_text.py
rmarrowstone: I will _try_ that. But I confess this comment is really a stand-in for my not fully understanding the relevance of the code we have for handling unicode if we're only supporting python3. All Python3 strings are already unicode, so I'm not sure what we need to worry about.

I _think_ the answer is that Ion supports some characters that are outside the range of valid UTF-8. I would really appreciate any context you had here.

amazon/ion/reader.py
rmarrowstone: Honestly I don't know. I added those comments only to hopefully get some input from others and/or as a note for myself to follow up later.

amazon/ion/writer_text.py
rmarrowstone: After analyzing the code this is _not_ used for text. It is used for numbers and timestamps. The nice part about `encode()` is that will error if the codepoints cannot be encoded in the encoding.

I'm going to set it to UTF-8 with a comment. Likewise for above. Thanks for checking me on this.

amazon/ion/reader.py
rmarrowstone: Yes. I see that this is used for both binary and text. (I overthought it at first).

amazon/ion/writer_text.py
rmarrowstone: Sorry I didn't see your reply before re-replying. Will follow up with you offline to clarify my understanding further.

PR #221, opened 2022-12-08
Remove Python 3.6 from Builds, Bump Minor Version

rmarrowstone:
This change move the oldest supported version of Python to 3.7.
Python 3.6 is now a year past EOL and is no longer supported on the
ubuntu images used for github builds.

This also bumps the *minor* version number. While there are no current
changes that _shouldn't_ work on 3.6, future changes may not.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ for Python.
 [![Build Status](https://travis-ci.org/amzn/ion-python.svg?branch=master)](https://travis-ci.org/amzn/ion-python)
 [![Documentation Status](https://readthedocs.org/projects/ion-python/badge/?version=latest)](https://ion-python.readthedocs.io/en/latest/?badge=latest)
 
-This package is designed to work with **Python 3.6+**
+This package is designed to work with **Python 3.6** and above. Sunsetted versions may not be supported once they are 6 months or more past EOL.
 
 ## Getting Started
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -5,7 +5,7 @@ for Python.
 [![Build Status](https://travis-ci.org/amzn/ion-python.svg?branch=master)](https://travis-ci.org/amzn/ion-python)
 [![Documentation Status](https://readthedocs.org/projects/ion-python/badge/?version=latest)](https://ion-python.readthedocs.io/en/latest/?badge=latest)
 
-This package is designed to work with **Python 3.6+**
+This package is designed to work with **Python 3.6** and above. Sunsetted versions may not be supported once they are 6 months or more past EOL.
```
tgregg, line 8: Should we update this to 3.7?

setup.py
```diff
@@ -51,7 +51,7 @@ def run_setup():
 
     setup(
         name='amazon.ion',
-        version='0.9.3',
+        version='0.10.3',
```
tgregg, line 54: Do we go to `0.10.1`?

rmarrowstone: facepalm. yes.

rmarrowstone: I even reviewed that and was like "yea looks right"

README.md
rmarrowstone: Good Question! I _did_ but then I reconsidered. It is still true that it was designed to work with Python 3.6, that hasn't changed. But there is nothing in this change that means it won't work with 3.6, and it's not true that it was designed to work with a more recent Python!

Maybe better would be to change it to:
> This package is designed to work with Python 3. <<short but descriptive statement about range of minor versions supported and api optimized for here that I'm still working on>>

tgregg: Works for me!

cheqianh (approved):

tgregg (approved):

PR #237, opened 2023-02-08
Adds format option for read command.

cheqianh:
#### Description:

This PR adds support for read format option - https://github.com/amazon-ion/ion-python/issues/234. Currently it only converts formats between Ion binary and Ion text. 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.6', '3.7', '3.8', '3.9', '3.10', 'pypy-3.6', 'pypy-3.7', 'pypy-3.8' ]
+        python-version: ['3.7', '3.8', '3.9', '3.10', 'pypy-3.7', 'pypy-3.8' ]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -344,6 +344,22 @@ def reset_for_each_execution(each_option):
     return api, format_option
 
 
+def generate_ion_file_with_format(file, binary):
+    cf = open("converted_file", "wb")
+    with open(file, "br") as fp:
+        ion.dump(ion.load(fp, single_value=False), cf, binary=binary)
+    return cf.name
```
tgregg, line 351: We should do nothing if the file is already in the requested format. That way we make sure to preserve e.g. whitespace in text and symbol table boundaries in binary.

cheqianh: I tried ask the tool to check file format by suffix (for ion_text) and IVM (for binary). However, suffix restriction is not supported in python-benchmark-cli yet since the tool only supports ion text and binary at the moment. 

The next step is to add a file format check mechanism and I'm going to learn from ion-java-benchmark-cli - check file header (E.g., an IVM), if it doesn't exist then fall back to suffix.

PR #243, opened 2023-03-10
Adds JSON/CBOR support and an Io-type option

cheqianh:
# Description: 
This PR integrates JSON/CBOR into the benchmark CLI. It additionally adds an IO type option to decide which IO type of the Ion data will be benchmarked. 

# Details

In details, this PR adds below features/components
1. Enable the CLI tool to benchmark Ion data as well as modern JSON/CBOR libraries concurrently.
2. Adds an IO type option to allow the tool benchmark time for either read/write in-memory buffers or files.
3. A bunch of refactor and unit testing.

# Output Example
## Example One 
Benchmark load API for all JSON libraries as well as ION

**Command and Output:**
```
python amazon/ionbenchmark/ion_benchmark_cli.py read --format json --format orjson --format simplejson --format ion_text --format ion_binary --format ujson
```
Screenshot 2023-03-09 at 5 53 25 PM

## Example Two 
Benchmark dumps API (--io-types buffer) for CBOR2,  and Ion_binary

**Command and Output:**
```
python amazon/ionbenchmark/ion_benchmark_cli.py write --io-type buffer --format cbor2 --format ion_binary test_unit_int
```
Screenshot 2023-03-09 at 5 56 19 PM

## Example Three  
Benchmark dump (--io-types file) API for CBOR2,  and Ion_binary

**Command and Output:**
```
python amazon/ionbenchmark/ion_benchmark_cli.py write --io-type file --format cbor2 --format ion_binary test_unit_int
```
Screenshot 2023-03-09 at 5 56 58 PM

# Follow-up issues
1. **Memory profiling enhancement**  
Currently, the tool only profile a memory usage peak. Due to the potential optimization Python may do, the memory usage metrics might be inaccurate after running different formats multiple execution times. Need more investigation for memory related metrics - https://github.com/amazon-ion/ion-python/issues/245
2. **Needs a tool to convert different file formats.** 
https://github.com/amazon-ion/ion-python/issues/234
3. **Needs pretty printed format options**
Right now, the options output in the output table is unclear. For example, (simpleion, ion_binary, file) represents that `--api` is simpleion, `--format` is ion_binary and `--io-type` is file. This needs improvement - https://github.com/amazon-ion/ion-python/issues/244. 

### DIFF
--- a/amazon/ionbenchmark/ion_benchmark_cli.py
+++ b/amazon/ionbenchmark/ion_benchmark_cli.py
@@ -46,8 +46,9 @@ Options:
      -c --c-extension <bool>            If the C extension is enabled, note that it only applies to simpleIon module.
                                         [default: True]
 
-     -f --format <format>               Format to benchmark, from the set (ion_binary | ion_text). May be specified
-                                        multiple times to compare different formats. [default: ion_binary]
+     -f --format <format>               Format to benchmark, from the set (ion_binary | ion_text | json | simplejson |
+                                        ujson | rapidjson | orjson | cbor | cbor2). May be specified multiple times to
+                                        compare different formats. [default: ion_binary]
 
      -p --profile                       (NOT SUPPORTED YET) Initiates a single iteration that repeats indefinitely until
                                         terminated, allowing users to attach profiling tools. If this option is
--- a/amazon/ionbenchmark/ion_benchmark_cli.py
+++ b/amazon/ionbenchmark/ion_benchmark_cli.py
@@ -47,7 +47,7 @@ Options:
                                         [default: True]
 
      -f --format <format>               Format to benchmark, from the set (ion_binary | ion_text | json | simplejson |
-                                        ujson | rapidjson | orjson | cbor | cbor2). May be specified multiple times to
+                                        ujson | rapidjson | cbor | cbor2). May be specified multiple times to
                                         compare different formats. [default: ion_binary]
 
      -p --profile                       (NOT SUPPORTED YET) Initiates a single iteration that repeats indefinitely until
@@ -94,9 +94,6 @@ from amazon.ionbenchmark.Io_type import Io_type
 pypy = platform.python_implementation() == 'PyPy'
 if not pypy:
     import tracemalloc
-    import orjson
-else:
-    import json as orjson
 
 BYTES_TO_MB = 1024 * 1024
 _IVM = b'\xE0\x01\x00\xEA'
@@ -112,7 +109,6 @@ output_file = 'dump_output'
 # Generates benchmark code for json/cbor/Ion load/loads APIs
 def generate_read_test_code(file, memory_profiling, format_option, binary, io_type, iterator=False, single_value=False,
                             emit_bare_values=False):
-    # if format_option == Format.ION_TEXT.value or format_option == Format.ION_BINARY.value:
     if format_is_ion(format_option):
         if io_type == Io_type.BUFFER.value:
             with open(file, "br") as fp:
@@ -207,9 +203,6 @@ def generate_read_test_code(file, memory_profiling, format_option, binary, io_ty
         benchmark_api = ujson.loads if io_type == Io_type.BUFFER.value else ujson.load
     elif format_option == Format.RAPIDJSON.value:
         benchmark_api = rapidjson.loads if io_type == Io_type.BUFFER.value else rapidjson.load
-    elif format_option == Format.ORJSON.value:
-        # orjson doesn't provide load API, so use loads for both file and buffer io_types.
-        benchmark_api = orjson.loads
     elif format_option == Format.CBOR.value:
         benchmark_api = cbor.loads if io_type == Io_type.BUFFER.value else cbor.load
     elif format_option == Format.CBOR2.value:
@@ -229,21 +222,6 @@ def generate_read_test_code(file, memory_profiling, format_option, binary, io_ty
             def test_func():
                 tracemalloc.start()
                 data = benchmark_api(benchmark_data)
-                global read_memory_usage_peak
-                read_memory_usage_peak = tracemalloc.get_traced_memory()[1] / BYTES_TO_MB
-                tracemalloc.stop()
-                return data
-    elif format_option == Format.ORJSON.value:
-        if not memory_profiling:
-            def test_func():
-                with open(file, 'br') as benchmark_file:
-                    data = benchmark_api(benchmark_file.read())
-                return data
-        else:
-            def test_func():
-                tracemalloc.start()
-                with open(file, 'br') as benchmark_file:
-                    data = benchmark_api(benchmark_file.read())
                 global read_memory_usage_peak
                 read_memory_usage_peak = tracemalloc.get_traced_memory()[1] / BYTES_TO_MB
                 tracemalloc.stop()
@@ -273,17 +251,21 @@ def generate_event_test_code(file):
 
 
 # Generates setup code for json/cbor benchmark code
-def generate_setup(format_option, c_extension=False, gc=False):
+def generate_setup(format_option, c_extension=False, gc=False, memory_profiling=False):
     if format_is_ion(format_option):
-        rtn = f'import amazon.ion.simpleion as ion;from amazon.ion.simple_types import IonPySymbol; ion.c_ext = ' \
-              f'{c_extension}; import tracemalloc'
+        rtn = f'import gc; import amazon.ion.simpleion as ion;from amazon.ion.simple_types import IonPySymbol; ' \
+              f'ion.c_ext = {c_extension}'
     else:
-        rtn = 'import tracemalloc'
+        rtn = 'import gc'
 
     if gc:
-        rtn += '; import gc; gc.enable()'
+        rtn += '; gc.enable()'
     else:
-        rtn += '; import gc; gc.disable()'
+        rtn += '; gc.disable()'
+
+    if memory_profiling:
+        rtn += '; import tracemalloc'
+
     return rtn
 
 
@@ -297,7 +279,7 @@ def read_micro_benchmark(iterations, warmups, c_extension, file, memory_profilin
                          iterator=False):
     file_size = Path(file).stat().st_size / BYTES_TO_MB
 
-    setup_with_gc = generate_setup(format_option=format_option, gc=False)
+    setup_with_gc = generate_setup(format_option=format_option, gc=False, memory_profiling=memory_profiling)
 
     test_code = generate_read_test_code(file, memory_profiling=memory_profiling,
                                         format_option=format_option, io_type=io_type, binary=binary)
@@ -316,7 +298,8 @@ def read_micro_benchmark_simpleion(iterations, warmups, c_extension, file, memor
                                    io_type, iterator=False):
     file_size = Path(file).stat().st_size / BYTES_TO_MB
 
-    setup_with_gc = generate_setup(format_option=format_option, c_extension=c_extension, gc=False)
+    setup_with_gc = generate_setup(format_option=format_option, c_extension=c_extension, gc=False,
+                                   memory_profiling=memory_profiling)
 
     test_code = generate_read_test_code(file, format_option=format_option, emit_bare_values=False,
                                         memory_profiling=memory_profiling, iterator=iterator, io_type=io_type,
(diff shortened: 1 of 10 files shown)


cheqianh:
The CI/CD failed because pypy does not support CBOR2.

cheqianh (commented):
The CI/CD failed because of that `pypy` does not support `cbor2`. We need to either,
(1) Disable all cbor2 feature for `pypy`, or
(2) treat `cbor2` as `cbor` for `pypy` environment

cheqianh (commented):
CI/CD failed due some json libraries does not support `pypy`. E.g., orjson

Without that, it works as expected with known 10 failed tests - https://github.com/amazon-ion/ion-python/issues/249

tests/test_benchmark_cli.py
```diff
@@ -135,39 +136,87 @@ def gather_all_options_in_list(table):
...
+)
+def test_read_json_format(f):
+    table = execution_with_command(['read', generate_test_path('integers.ion'), '--format', f'{f}'])
+    assert gather_all_options_in_list(table) == sorted([('simple_ion', f'{f}', 'file')])
+
+
+@parametrize(
+    *tuple((f.value for f in Format.Format if Format.format_is_json(f.value)))
+)
+def test_write_json_format(f):
```
linlin-s, line 196: Should we also include the test case for `test_write_cbor_format` here?

tgregg:
> CI/CD failed due some json libraries does not support `pypy`. E.g., orjson. We can either
> 
>     1. Treat `orjson` as regular built-in json library when using `pypy` interpreter, or
> 
>     2. Remove orjson support on pypy.
> 
> 
> Without that, it works as expected with known 10 failed tests - #249

Let's go with option 2

ion-c
tgregg, line 1: Since this introduces test failures that haven't yet been resolved, let's leave the ion-c submodule update for a separate PR, and revert it here.

amazon/ionbenchmark/ion_benchmark_cli.py
tgregg, line 1: For the `--api` option, the allowed values should be changed to something like `load_dump` and `streaming`. The option `simple_ion` is no longer accurate for the other formats, and looks weird in the options list when those formats are selected.

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -84,52 +100,169 @@
...
+                                break
+                    global read_memory_usage_peak
+                    read_memory_usage_peak = tracemalloc.get_traced_memory()[1] / BYTES_TO_MB
+                    tracemalloc.stop()
+                    return it
+    return test_func
+
+
+# Generates benchmark code for json/cbor load/loads APIs
+def generate_read_test_code(file, memory_profiling, format_option, binary, io_type):
```
tgregg, line 200: Is there a reason we can't follow the same path for Ion?

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -264,33 +431,126 @@ def write_micro_benchmark_simpleion(iterations, warmups, c_extension, obj, file,
...
+    timeit.timeit(stmt=test_func, setup=setup_with_gc, number=warmups)
+
+    # iteration
+    result_with_gc = timeit.timeit(stmt=test_func, setup=setup_with_gc, number=iterations) / iterations
+
+    return file_size, result_with_gc
+
+
+# Generates benchmark code for json dump API
+def generate_write_test_code(obj, memory_profiling, format_option, io_type, binary):
```
tgregg, line 454: Is there a reason Ion can't follow this path?

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -360,55 +646,50 @@ def ion_python_benchmark_cli(arguments):
...
-
+    # initialize benchmark report table
+    table = identify_report_table(command)
+
+    for each_option in option_configuration_combination:
+        print(f'Generating option {each_option}...')
+        # reset each option configuration
+        api, format_option, io_type = reset_for_each_execution(each_option)
+        binary = format_is_binary(format_option)
+        # TODO. currently, we must provide the tool a corresponding file format for read benchmarking. For example,
```
tgregg, line 662: The tool *to convert to* a corresponding file format for read benchmarking. ?

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -84,52 +100,169 @@
...
-                    it = ion.load(fp, single_value=single_value, emit_bare_values=emit_bare_values,
-                                  parse_eagerly=False)
+            if not iterator:
+                def test_func():
+                    tracemalloc.start()
+                    data = ion.loads(benchmark_data, single_value=single_value, emit_bare_values=emit_bare_values)
+                    global read_memory_usage_peak
+                    read_memory_usage_peak = tracemalloc.get_traced_memory()[1] / BYTES_TO_MB
+                    tracemalloc.stop()
+                    return data
```
tgregg, line 138: `test_func()` is what is being timed, correct? That means the cost of using `tracemalloc` is included in the results. Is there a way for us to extract this logic outside of the timed block?

tgregg (commented):
Since format conversion is left as a TODO, can you explain what is actually happening today in each of your examples in the description? Do those come from an actual execution?

Once we have data conversion, I'd expect the data size stat to be different for each of the formats, as it will reflect the size of the *converted* data, for both read and write.

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -55,24 +55,40 @@
                                         error will be raised if this option is used when multiple values are specified
                                         for other options. Not enabled by default.
 
+     -i --io-type <io_type>             The source or destination type, from the set (buffer | file). If buffer is
+                                        selected, buffers the input data in memory before reading and writes the output
+                                        data to an in-memory buffer instead of a file. [default: file]
+
```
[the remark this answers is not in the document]
cheqianh: Thank you catching this, added.

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -318,8 +578,7 @@ def insert_into_report_table(table, row):
 def identify_report_table(command):
     if command == 'read':
         return identify_report_table_helper(
-            ['file_size (MB)', 'options', 'total_time (s)', 'conversion_\ntime (s)', 'conversion_time/\ntotal_time (%)',
-             'memory_usage_peak (MB)'])
+            ['file_size (MB)', 'options', 'total_time (s)', 'memory_usage_peak (MB)'])
```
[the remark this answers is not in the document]
cheqianh: Yep, we should unify the read/write tables ultimately. I left it separately for now because Read/Write command may generate different metrics. 

One example is the previous `READ` table, which includes an additional `conversion time` column to indicate the execution time of whether raw Python objects are wrapped into simpleIon types.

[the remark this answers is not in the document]
linlin-s: Thanks for explaining.

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -84,52 +100,169 @@
...
+                    tracemalloc.stop()
+                    return it
     else:
-        if not iterator:
+        if not memory_profiling:
+            if not iterator:
+                def test_func():
+                    with open(file, "br") as fp:
+                        data = ion.load(fp, single_value=single_value, emit_bare_values=emit_bare_values)
+                    return data
```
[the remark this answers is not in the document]
cheqianh: > if that's possible to merge the similar logic in two different conditions into one reusable function?

Good question. We may not want to merge them into one function now but it's possible to refactor some logics if `test_func()` becomes complex later.

This is because the whole functionality of the `test_func` will be benchmarked and they (`load` and `loads`) are different APIs with a similar signature.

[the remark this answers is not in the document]
cheqianh: > if that's possible to merge the similar logic in two different conditions into one reusable function?

Good question. We may not want to merge them into one function now but it's possible to refactor some logics if `test_func()` becomes complex later.

This is because the whole functionality of the `test_func` will be benchmarked and they (`load` and `loads`) are different APIs with a similar signature.

tests/test_benchmark_cli.py
cheqianh: [87e7f65](https://github.com/amazon-ion/ion-python/pull/243/commits/87e7f65f9d05fc527820226a5dc93c9ee1d22954) addressed them.

cheqianh:
> Since format conversion is left as a TODO, can you explain what is actually happening today in each of your examples in the description? Do those come from an actual execution?

Yep, these metrics are generated by the tool directly. Overall, the tool is assuming that we had the format conversion functionality already but it doesn't do any conversion work for files. 

For each example,
(1) Benchmarking both read and write for JSON/Ion_text doesn't require format conversion. One thing I want to point out is that ion_binary is compared but it's doing the exact same thing as format `ion_text`. We had an enhancement ticket describing this - https://github.com/amazon-ion/ion-python/issues/234. Once we supported format conversion, the tool will update it automatically without any changes. 

(2) and (3)
Benchmarking `write` command for CBOR/Ion doesn't require format conversion; we just need to provide python layer objects and doesn't need to care about how they encode the objects and write them into files. But I found one thing that breaks the fairness of Ion/CBOR benchamrking is the slight difference in python objects generated by them (E.g., cbor2 reads "9233720363654371807" as `-12852` in python). But the tool will automatically generate the correct object once we support the format conversion. (This is the tenet when I work on this PR; I always assume that we had the formats conversion tool already so `cbor2` will generate correct `cbor2` objects from the desired file format)

> Once we have data conversion, I'd expect the data size stat to be different for each of the formats, as it will reflect the size of the converted data, for both read and write.

Yep, the [converted file](https://github.com/amazon-ion/ion-python/pull/243/files#diff-193f9a476a85a550a2aa80e3a9ad54272c2661544bc96455fd6439664e1b9df1R666) is being used for calculating the file size. Once we support the formats conversion, it will generate new file size automatically.

amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -94,6 +94,9 @@
 pypy = platform.python_implementation() == 'PyPy'
 if not pypy:
     import tracemalloc
+    import orjson
+else:
+    import json as orjson
```
tgregg, line 99: I think we should just throw if `orjson` is requested when using pypy, rather than silently replacing it.

cheqianh: Yeah, I try to push other comments first to see if it pass the CI/CD, still working on it.

amazon/ionbenchmark/ion_benchmark_cli.py
cheqianh: Yes, we execute it twice; once with `memory_profiling` enabled, and once with it disabled. 

We profile the memory before all performance benchmarking [here](https://github.com/amazon-ion/ion-python/blob/master/amazon/ionbenchmark/ion_benchmark_cli.py#L284-L287). Then we [benchmark](https://github.com/amazon-ion/ion-python/blob/master/amazon/ionbenchmark/ion_benchmark_cli.py#L289-L293) the performance separately.

amazon/ionbenchmark/ion_benchmark_cli.py
cheqianh: Fixed.

amazon/ionbenchmark/ion_benchmark_cli.py
cheqianh: same as above comment.

amazon/ionbenchmark/ion_benchmark_cli.py
cheqianh: Merged them into one.

amazon/ionbenchmark/ion_benchmark_cli.py
cheqianh: Changed.

cheqianh:
There are few `pypy` issue failing the CI/CD, working on it.

cheqianh:
The new commit addressed all the feedback above, passed all the CI/CD and `pypy` incompatible issue, and deprecated `orjson` for now. 

Here is the GH [issue](https://github.com/amazon-ion/ion-python/issues/253) to add orjson back.

tgregg (approved):

PR #246, opened 2023-03-13
Adds the simpleEnum class back.

cheqianh:
*Description of changes:*
* Adds its meta class back without using six.
* Adds unit tests back
* Marks it as deprecation. Adds deprecation warnings for the class.

*Issue #, if available:*
https://github.com/amazon-ion/ion-python/issues/241

### DIFF
--- a/amazon/ion/util.py
+++ b/amazon/ion/util.py
@@ -17,10 +17,14 @@
 import sys
 
 from collections import namedtuple
+from warnings import warn
+
+import six
 
 
 class _RecordMetaClass(type):
     """Metaclass for defining named-tuple based immutable record types."""
+
     def __new__(cls, name, bases, attrs):
         if attrs.get('_record_sentinel') is None:
             field_declarations = []
@@ -78,6 +82,7 @@ def record(*fields):
     Args:
         fields (list[str | (str, any)]): A sequence of str or pairs that
     """
+
     class RecordType(object, metaclass=_RecordMetaClass):
         _record_sentinel = True
         _record_fields = fields
@@ -94,12 +99,14 @@ def coroutine(func):
     Returns:
         Callable: The decorated generator.
     """
+
     def wrapper(*args, **kwargs):
         gen = func(*args, **kwargs)
         val = next(gen)
         if val != None:
             raise TypeError('Unexpected value from start of coroutine')
         return gen
+
     wrapper.__name__ = func.__name__
     wrapper.__doc__ = func.__doc__
     return wrapper
@@ -142,6 +149,7 @@ class CodePoint(int):
     """Evaluates as the ordinal of a code point, while also containing the unicode character representation and
     indicating whether the code point was escaped.
     """
+
     def __init__(self, *args, **kwargs):
         self.char = None
         self.is_escaped = False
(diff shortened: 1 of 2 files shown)


amazon/ion/util.py
```diff
@@ -217,15 +226,106 @@ def bit_length(value):
...
+        """Looks up an enumeration value field by integer value."""
+        return cls._enum_members[name]
+
+    def __iter__(self):
+        """Iterates through the values of the enumeration in no specific order."""
+        return iter(self._enum_members.values())
+
+
+class Enum(int, metaclass=_EnumMetaClass):
+    """Simple integer based enumeration type.
```
popematt, line 274:
```suggestion
    """This is a deprecated, internal-only class in ion-python—do NOT use it for any reason.
    
    Simple integer based enumeration type.
```

amazon/ion/util.py
```diff
@@ -217,15 +226,106 @@ def bit_length(value):
...
     def total_seconds(td):
         return td.total_seconds()
 
-
 bit_length.__doc__ = 'Returns the bit length of an integer'
 total_seconds.__doc__ = 'Timedelta ``total_seconds`` with backported support in Python 2.6'
+
+
+class _EnumMetaClass(type):
+    """Metaclass for simple enumerations.
```
popematt, line 243: Let's make it obvious in the doc string too.
```suggestion
    """This is a deprecated, internal-only class in ion-python—do NOT use it for any reason.

    Metaclass for simple enumerations.
```

cheqianh:
The CI/CD failed due to C extension. which does not have any change. Looking into it right now.

cheqianh:
Found the root, see - https://github.com/amazon-ion/ion-python/issues/249

We will skip these 10 unit tests now.

amazon/ion/util.py
```diff
@@ -217,15 +226,106 @@ def bit_length(value):
...
+    _enum_members = {}
+
+    def __new__(cls, name, value):
+        return int.__new__(cls, value)
+
+    def __init__(self, name, value):
+        warn(f'{self.__class__.__name__} was deprecated in favor of the deprecation of `amazon.ion.Enum`. However, for '
+             f'compatibility reasons, `amazon.ion.Enum` has been placed here. We recommend considering using `IntEnum` '
+             f'as an alternative.',
+             DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
cheqianh: Agree, changed.

amazon/ion/util.py
cheqianh: Changed.

amazon/ion/util.py
cheqianh: Addressed.

popematt:
Looks like the build is failing in the Python 3.10 workflow because of an error while trying to load the c extension. Do you know why that might be?

cheqianh:
The CI/CD will fail for all python 3.x versions due to the recent ion-c release. I've navigated [the commit](https://github.com/amazon-ion/ion-c/commit/621e3796968170c520ea06a89b4baea26c43b070) that causes this change. Further investigation is required to identify the root cause. 

Reference issue - https://github.com/amazon-ion/ion-python/issues/249 for details.

ion-c
tgregg, line 1: Let's revert the submodule update and address that in a separate PR, since it introduced errors that haven't yet been diagnosed.

cheqianh:
The CI/CD passed due to the mitigation we had https://github.com/amazon-ion/ion-python/pull/250, which the commit will be reverted once the root cause is fixed.

popematt (approved):

PR #261, opened 2023-04-08
fix: seek to previous position instead of 0 after reading IVM

melsabagh:
*Issue #, if available:* https://github.com/amazon-ion/ion-python/issues/251

*Description of changes:* Remember current file position then seek back to it after reading the IVM instead of blindly seeking to position 0.


popematt (approved):
Thanks for the fix!

PR #262, opened 2023-04-14
Replace Usages of record with NamedTuple

rmarrowstone:
This change replaces all of the usages of the record super-type
constructor with the built-in NamedTuple. This makes the code easier
to read (no custom special type mojo) and enables IDEs and other tooling
to provide type hints when constructing and field name completion when
using instances.

The types which extend NamedTuple are immutable as they were before.
I'm sure there was a reason record was created initially, but I don't
see a good reason to keep it now, particularly given the benefits of
just using NamedTuple.

I kept the record constructor and metaclass to avoid breaking a
consumer who is using them. We can remove them when we release an MV.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -12,8 +12,10 @@
 # specific language governing permissions and limitations under the
 # License.
 
-"""Ion core types."""
 from enum import IntEnum
+from typing import NamedTuple, Optional, Any, Union, Sequence, Coroutine
+
+from amazon.ion.symbols import SymbolToken
 
 # in Python 3.10, abstract collections have moved into their own module
 # for compatibility with 3.10+, first try imports from the new location
@@ -27,8 +29,6 @@ except:
 from datetime import datetime, timedelta, tzinfo
 from decimal import Decimal, ROUND_FLOOR, Context, Inexact
 from math import isnan
-
-from .util import record
 
 
 class IonType(IntEnum):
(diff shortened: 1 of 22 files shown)


cheqianh (approved):
Thanks for the change! **I saw a performance improvement for the object creation (specifically 21.8% for IonPySymbol).** 

Just saw something about [creating Python immutable objects](https://stackoverflow.com/a/58992808). Since Python 3.7, users can create immutable objects by setting [dataclass `frozen=True`](https://docs.python.org/3/library/dataclasses.html#frozen-instances).

The creation and initialization process of IonPyTypes, such as IonPySymbol, with a metaclass of `recordMetaClass` can be extremely long, which, as a result, significantly reducing serialization speed. I believe this PR will mitigate the expensive cost of object initialization.

After replacing `record` with `namedtuple`, there is a consistent performance improvement in object initialization. see the benchmark script below,
```.py
start = time.time()
# Simply execute it for one million times for better read experience 
for i in range(1000000):
    a = IonPySymbol('text', 2)
print('------------- IonPySymbol', time.time() - start)
```
and get
```
# percentage increase = before - after / before = 21.8%

# Before: 
------------- IonPySymbol 1.1837401390075684
# After:
------------- IonPySymbol 0.9258627891540527
```

I wrote a custom mutable symbol-token-like object and saw a 4x improvement (1.2->0.3s / 1m executions). It would be great to figure out the tradeoffs of making these objects immutable as well. I'm going to put together some POC for performance topic later 🐮 and this PR would be a good call out.

rmarrowstone:
Thanks for the profiling deep-dive Eric! I had looked at parsing performance and timing of the test suite, but only anecdotally. It seemed that there may have been improvement, there certainly wasn't regression but it was far from scientific or statistically meaningful.

I have a small fixup to a type hint that I want to post before merging. Expect that shortly.

PR #267, opened 2023-05-12
Binary Reader Performance Refactor

rmarrowstone:
This commit refactors the pure python (non-extension) binary reader to
improve performance. My testing shows it to be roughly 2.5x faster.

It ditches the coroutine dispatching overhead within the reader and
replaces the stateful buffer with an immutable buffer that uses
memoryviews to decrease memory allocations. The immutability of the
buffer is less important for the binary reader but will be important
for the text reader where we often need to look ahead.

It contains a change to the IonThunkEvent that should have a positive
impact on the text reader as well. More impactful will be to apply
this pattern to the text reader.

I left the managed_reader unchanged though de-coroutining that and
minimizing managed event construction could yield more gains for both
text and binary.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -276,7 +276,7 @@ class IonThunkEvent(IonEvent):
 
     The `value` will be materialized on first access and cached.
 
-    Accessing the value by it's slot: ``event[2]`` will avoid materialization.
+    Accessing the value by its slot: ``event[2]`` will avoid materialization.
     """
 
     @property
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -11,7 +11,6 @@
 # OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the
 # License.
-
 from enum import IntEnum
 from typing import NamedTuple, Optional, Any, Union, Sequence, Coroutine
 
@@ -272,21 +271,20 @@ class MemoizingThunk(object):
 
 
 class IonThunkEvent(IonEvent):
-    """An :class:`IonEvent` whose ``value`` field is a thunk."""
-    def __new__(cls, *args, **kwargs):
-        if len(args) >= 3:
-            args = list(args)
-            args[2] = MemoizingThunk(args[2])
-        else:
-            value = kwargs.get('value')
-            if value is not None:
-                kwargs['value'] = MemoizingThunk(kwargs['value'])
-        return super(IonThunkEvent, cls).__new__(cls, *args, **kwargs)
+    """
+    A lazy `IonEvent` whose ``value`` field is a thunk.
+
+    The `value` will be materialized on first access and cached.
+
+    Accessing the value by it's slot: ``event[2]`` will avoid materialization.
+    """
 
     @property
     def value(self):
-        # We're masking the value field, this gets around that.
-        return self[2]()
+        if hasattr(self, 'cached_value'):
+            return self.cached_value
+        self.cached_value = self[2]()
+        return self.cached_value
 
 # Singletons for structural events
 ION_STREAM_END_EVENT = IonEvent(IonEventType.STREAM_END)
@@ -314,11 +312,11 @@ class Transition(NamedTuple):
     This is generally used as a result of a state-machine.
 
     Args:
-        event (Optional[DataEvent]): The event associated with the transition.
+        event (Union[DataEvent, IonEvent, None]): The event associated with the transition.
         delegate (Coroutine): The co-routine delegate which can be the same routine from
             whence this transition came.
     """
-    event: Optional[DataEvent]
+    event: Union[DataEvent, IonEvent, None]
     delegate: Coroutine
 
 
--- a/amazon/ion/reader_binary.py
+++ b/amazon/ion/reader_binary.py
@@ -11,21 +11,22 @@
 # OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the
 # License.
-
+from collections import deque
 from datetime import timedelta
 from decimal import Decimal, localcontext
 from enum import IntEnum
 from functools import partial
 from io import BytesIO
 from struct import unpack
-from typing import NamedTuple, Optional, Sequence, Coroutine
+from typing import NamedTuple, Optional, Sequence, Callable, Tuple, List
 
-from .core import ION_STREAM_INCOMPLETE_EVENT, ION_STREAM_END_EVENT, ION_VERSION_MARKER_EVENT,\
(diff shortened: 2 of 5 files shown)


amazon/ion/reader_binary.py
```diff
@@ -208,406 +208,275 @@ def _parse_sid_iter(data):
...
         else:
-            field_name = None
+            if read_event.type is ReadEventType.DATA:
+                raise TypeError("Next or Skip expected!")
+            skip_or_next = read_event.type
+
+        ion_event = None
+        if skip_or_next is ReadEventType.SKIP:
+            if parent_type is None:
+                raise TypeError("Cannot Skip Top-level values!")
```
tgregg, line 429: This message doesn't look quite right. You should be able to skip top-level values (e.g. a container at the top level). Is the problem that skip is being requested before `NEXT` to position the reader on the value?

tests/test_sliceable_buffer.py
```diff
@@ -0,0 +1,200 @@
...
+class _P(NamedTuple):
+    desc: str
+    actions: Sequence[Callable]
+    is_unicode: bool = False
+
+    def __str__(self):
+        return self.desc
+
+
+@parametrize(
```
tgregg, line 76: Let's add some parameters to test the expected `IncompleteReadError` behavior.

amazon/ion/sliceable_buffer.py
```diff
@@ -0,0 +1,164 @@
...
+        Skip max(n, size) bytes, return (skipped, new buffer).
+
+        Chunks which are no longer readable are dropped from the new buffer.
+        """
+        size = self.size
+        chunks = self._chunks
+        offset = self._offset
+        endpos = offset + n
+
+        if size <= n:
```
tgregg, line 134: Should `size < n` be `IncompleteReadError`?

amazon/ion/reader_binary.py
```diff
@@ -208,406 +208,275 @@ def _parse_sid_iter(data):
...
+
+            to_skip = limit - cursor
+            (skipped, buffer) = buffer.skip(to_skip)
+            cursor += skipped
+            if cursor < limit:
+                ion_event = ION_STREAM_INCOMPLETE_EVENT
```
tgregg, line 435: `IncompleteReadError` is used to convey this in the case of `read`. Should we be consistent?

tgregg (approved):
Faster and easier to understand. That's a win-win.

amazon/ion/reader_binary.py
rmarrowstone: I'm assuming that you get that the buffer talks in terms of the exceptions and the parser loop talks in terms of Ion Events and this comment is saying the same as above just at call-site. LMK if I'm misinterpreting.

amazon/ion/sliceable_buffer.py
rmarrowstone: I agree with your insight. However, allowing the partial skip keeps memory usage lower when skipping a large container. That may be a premature optimization, but it is how the existing skip works and I didn't want to introduce a potential regression. Let me add a comment there.

amazon/ion/core.py
```diff
@@ -272,21 +271,20 @@ def __repr__(self):
...
-            value = kwargs.get('value')
-            if value is not None:
-                kwargs['value'] = MemoizingThunk(kwargs['value'])
-        return super(IonThunkEvent, cls).__new__(cls, *args, **kwargs)
+    """
+    A lazy `IonEvent` whose ``value`` field is a thunk.
+
+    The `value` will be materialized on first access and cached.
+
+    Accessing the value by it's slot: ``event[2]`` will avoid materialization.
```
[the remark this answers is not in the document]
rmarrowstone: I'm just lousy with `it's` vs `its`.

amazon/ion/reader_binary.py
tgregg: Yes, it's the same comment as the above.

amazon/ion/sliceable_buffer.py
tgregg: I agree that partial skip should be allowed, but should it skip what it can and then throw `IncompleteReadError`?

rmarrowstone: Not really: if there are any changes to state they need to be returned from the call with a new buffer and not mutate the old one.

rmarrowstone:
I wanted to answer two offline questions about performance. One was about an observed difference in reported to actual perf. gains, the other was about how this compares to the C-extension parsing performance. I had been solely testing the streaming event API which has no C-extension implementation currently.

The results below are the average of 3 runs of parsing 10k records of binary telemetry log data.

Streaming Old: 14.7 seconds
Streaming New: 6.3 seconds

simpleion.load Old: 39.4 seconds
simpleion.load New: 20.1 seconds
simpleion.load C-ext: 3.0 seconds

So the c-extension for simpleion.loads is still significantly faster!

amazon/ion/reader_binary.py
rmarrowstone: Yea... I added the message only based on what the `reader_trampoline` says.

tests/test_sliceable_buffer.py
rmarrowstone: I think the `read_byte` as already covered... there's nothing really interesting there as the buffer is either empty or it's not. I added tests to better verify lack of mutation for `read_slice` and resulting buffer for partial skip.

amazon/ion/reader_binary.py
```diff
@@ -208,406 +208,275 @@ def _parse_sid_iter(data):
...
 
-@coroutine
-def _ivm_handler(ctx):
-    _, self = yield
+def _var_uint_field_handler(handler, context: _ParserContext):
+    """Handler co-routine for variable unsigned integer fields that.
+
+    Invokes the given ``handler`` function with the read field and context,
+    then immediately yields to the resulting co-routine.
+    """
```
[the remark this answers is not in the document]
rmarrowstone: Realized I missed this in my fix-up commit. Will fix!

cheqianh (approved):
LGTM. Thanks for putting this together!

amazon/ion/sliceable_buffer.py
```diff
@@ -0,0 +1,164 @@
...
+    def skip(self, n):
+        """
+        Skip max(n, size) bytes, return (skipped, new buffer).
+
+        Chunks which are no longer readable are dropped from the new buffer.
+        """
+        size = self.size
+        chunks = self._chunks
+        offset = self._offset
+        endpos = offset + n
```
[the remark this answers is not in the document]
rmarrowstone: 😨  good catch!

PR #268, opened 2023-05-16
Increases benchmark warm-up and decreases the threshold value

cheqianh:
This PR 
(1) Addressed a memory leak issue - [link](https://github.com/amazon-ion/ion-python/pull/268/files#diff-535412a8258ff8b32325ed24da1018a8b331ec818afbda63f3d585d631c5847dR1549)
(2) Experimented the GHA runners' resource limitation and switched to MacOS for 1k iterations and warmups - [link](https://github.com/amazon-ion/ion-python/pull/268/files#diff-b242f7745129c84b875ab3f6f31df4796e8bbe4eac635052f979798c1733b839R15)
(3) Added a method for Ion binary/text conversion - [link](https://github.com/amazon-ion/ion-python/pull/268/files#diff-110d5a7302a3f14125871477abbf6e9ae974075a61e3fbaf0bcb47bf2e575350R39-R69)
(4) Fixed a comparison result mismatch issue by sorting the combination list - [link](https://github.com/amazon-ion/ion-python/pull/268/files#diff-193f9a476a85a550a2aa80e3a9ad54272c2661544bc96455fd6439664e1b9df1R695)

The threshold is set to 0.2 for now, we will work on its enhancement in the future. The [latest performance detection](https://github.com/amazon-ion/ion-python/actions/runs/5204562879/jobs/9388937206) failed due to the variance.  


tgregg:
A threshold of 0.5 still allows a 50% performance regression, right? That's still not going to be low enough for our needs. How did you select 150 for warmups/iterations? Are you seeing diminishing returns in terms of reduced variance as you increase the warmups/iterations, or does the regression detection workflow just start taking too long?

cheqianh:
If 0.5 doesn't satisfy our needs, we need to consider other improvements as well. 

The Github runner only allocate fixed resource (e.g., memory) to jobs. Running over 200 times make the job cancel automatically. Additionally the relative-difference is usually < 0.3 but sometimes one test (real_worlds_data_1) fail with 0.5+ perf regressions.   

One possible solution to enable more iteration and warmups are using a larger runner - https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

cheqianh:
If 0.5 doesn't satisfy our needs, we need to consider other improvements as well. 

The Github runner only allocate fixed resource (e.g., memory) to jobs. Running over 200 times make the job cancel automatically. Additionally the results of relative-difference are usually < 0.3 but sometimes one test (real_worlds_data_1) fail with 0.5+ perf regressions (and if the workflow fails, it's always this file).   

One possible solution to enable more iteration and warmups is using a larger runner - https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

cheqianh:
Found a small memory leak in each simpleion.dumps(), which causes the workflow exceeds the resource limit. Investigating the root.

cheqianh:
Confirmed that `execution_with_command(['read', file, '--api', 'load_dump', '--api', 'streaming'])` has memory leak for simpleion `load` api and `testStructs.10n`. Didn't see any memory leak for `dump` API.

.github/workflows/performance-regression.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Detect Regression
     needs: PR-Content-Check
     if: ${{ needs.PR-Content-Check.outputs.result == 'pass' }}
-    runs-on: ubuntu-latest
+    runs-on: macos-latest # ubuntu-latest
```
popematt, line 15: In general, we should only use MacOS runners if we need to test specifically that something works correctly on MacOS or we're distributing a MacOS-specific binary.

What's the rationale for running on MacOS here?

cheqianh: I'm trying to run more warm-ups and iterations for the benchmark-cli command in order to generate a consistent threshold value to help us identify the performance regression. However, command with 150+ executions will hit the resource limitation of the GHA linux runner.

popematt: Which resource limit? Is it causing the job to timeout?

cheqianh: https://github.com/actions/runner-images/issues/6680

cheqianh: Also I found another issue that generates the wrong threshold value so it's not stable.

cheqianh:
4.55e-03 and 4.10e-03 for sexp_write

cheqianh:
This PR 
(1) Addressed a memory leak issue - []
(2) Experimented the GHA runners' resource limitation and switched to MacOS for 1k iterations and warmups - []
(3) Added a method for Ion binary/text conversion - []
(4) Fixed a comparison result mismatch issue by sorting the combination list - []

The threshold is set to 0.2 for now, we will work on its enhancement in the future.

.github/workflows/performance-regression.yml
```diff
@@ -65,27 +65,30 @@ jobs:
...
+          python amazon/ionbenchmark/ion_benchmark_cli.py write -w $warmups --iterations $iterations -o /Users/runner/work/ion-python/ion-python/benchmarkResults/realWorldData02_write/new --io-type file --io-type buffer --format ion_text --format ion_binary --api load_dump /Users/runner/work/ion-python/ion-python/testData/realWorldData02.10n
+          python amazon/ionbenchmark/ion_benchmark_cli.py write -w $warmups --iterations $iterations -o /Users/runner/work/ion-python/ion-python/benchmarkResults/realWorldData03_write/new --io-type file --io-type buffer --format ion_text --format ion_binary --api load_dump /Users/runner/work/ion-python/ion-python/testData/realWorldData03.10n
 
       # Generates performance results for the previous commit
       - name: Checkout the master branch of the Ion Python
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.pull_request.head.sha }} # TODO change back to `master`
+          ref: ${{ github.event.pull_request.head.sha }} # TODO change back to master after
```
tgregg, line 91: after what?

.github/workflows/performance-regression.yml
```diff
@@ -97,21 +100,24 @@ jobs:
 
       - name: Running performance benchmark
         working-directory: ./ion-python-master
+        env:
+          warmups: 1000
```
tgregg, line 104: How long does 1000 warmups take? If it's longer than 30 seconds or so, that's probably enough time for us to achieve as much consistency as we're going to get. Can we shift some of these warmups to timed iterations?

amazon/ionbenchmark/Format.py
```diff
@@ -19,7 +36,37 @@ def format_is_binary(format_option):
...
+            with open(temp_file_name, 'bw') as fp:
+                if format_option == Format.ION_BINARY.value:
+                    simpleion.dump(obj, fp, binary=True)
+                else:
+                    simpleion.dump(obj, fp, binary=False)
+        else:
+            shutil.copy(file, temp_file_name)
+    else:
+        # Copy the file
+        shutil.copy(file, temp_file_name)
```
tgregg, line 67: For large files, always copying may become a burden, if no conversion was required.

tgregg (approved):

cheqianh:
Opened an issue for avoiding file copy - https://github.com/amazon-ion/ion-python/issues/269
Updated the issue for accurate threshold by using percentile - https://github.com/amazon-ion/ion-python/issues/265

amazon/ionbenchmark/Format.py
cheqianh: - https://github.com/amazon-ion/ion-python/issues/269

PR #271, opened 2023-07-05
Add module to load/dump standard protocol buffers

popematt:
**Issue #, if available:**

None

**Description of changes:**

Adds a module that provides functionality to make it easier to support standard protocol buffers in the benchmark CLI. Functionality includes:
* Wrapper for a protocol buffer `Message` that provides `load`/`loads` and `dump`/`dumps` for that message type
* Function to dynamically load a protocol buffer `Message` from a python module at runtime
* Function to dynamically generate a protocol buffer `Message` from a proto file descriptor at runtime

The functions for loading the `Message` are important because we can't hard-code the `Message` implementation into the benchmark CLI (since it's not necessarily known at compile time), so we need some way to load them at runtime.

### DIFF
--- a/amazon/ionbenchmark/proto.py
+++ b/amazon/ionbenchmark/proto.py
@@ -67,7 +67,7 @@ class ProtoSerde:
 
 def get_message_type_from_py(type_name, module_name, module_file=None):
     """
-    Import and load a protocol buffer message from a Python module at runtime.
+    Import and load a protocol buffer message class from a Python module at runtime.
 
     A Python protocol buffer module can be created using `protoc --python_out=...`.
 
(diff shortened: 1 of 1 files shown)


tgregg (approved):

amazon/ionbenchmark/proto.py
```diff
@@ -0,0 +1,128 @@
...
+    def dump(self, obj, fp):
+        """
+        Serialize a protocol buffer message to a file.
+        """
+        fp.write(obj.SerializeToString())
+
+
+def get_message_type_from_py(type_name, module_name, module_file=None):
+    """
+    Import and load a protocol buffer message from a Python module at runtime.
```
rmarrowstone, line 70: nit: load a protocol buffer message _type_ (or class?), right?

rmarrowstone (approved):

amazon/ionbenchmark/proto.py
popematt:
```suggestion
    Import and load a protocol buffer message class from a Python module at runtime.
```

PR #272, opened 2023-07-12
Remove '--io_type file' from performance regression workflow

popematt:
**Issue #, if available:**

None

**Description of changes:**

This should speed up the performance testing workflow by _at least_ one order of magnitude, and help contribute to the stability of the benchmark results, since the serde time is often (and especially with smaller data) dominated by the file system access time.


amazon/ionbenchmark/Command.py
```diff
@@ -1,8 +0,0 @@
-from enum import Enum
-
-
-class Command(Enum):
```
tgregg, line 4: Accidentally deleted?

amazon/ionbenchmark/Command.py
```diff
@@ -1,8 +0,0 @@
-from enum import Enum
```
cheqianh, line 1: QQ, do we still need COMMAND?

amazon/ionbenchmark/Command.py
popematt: Good catch.

It's such a simple PR, that I thought I could get away without checking myself before posting it. 😅
...but I was wrong.

cheqianh (commented):
Seems that CI/CD failed for PYPY due to deletion of command.

amazon/ionbenchmark/Command.py
popematt: This was an accidental deletion. I have re-added it.

cheqianh (approved):

PR #277, opened 2023-07-20
Enable GC For Repeated Runs

rmarrowstone:
In looking into why the pypy profiling was behaving erratically on GH
I observed that memory usage for the Ion Text tests in pure-python
mode was quite high: roughly 3gb for less than 500 iterations.

pypy uses the pure-python implementation.

In looking at the code I observed that setup_with_gc was not actually
being setup _with_ gc. Enabling it in these places makes a substantial
difference in memory use on my laptop.

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -33,7 +33,7 @@ from .writer import blocking_writer
 from .writer_binary import binary_writer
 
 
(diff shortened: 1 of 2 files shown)


rmarrowstone:
Some of the regression tests, but not all failed on cpython, which is unsurprising given that gc will add some time.

cheqianh (approved):

Issue #265, opened 2023-05-10
ion-python-benchmark-cli's threshold value need to be more accurate.


cheqianh:
The benchmark-cli's performance detection workflow fails if the performance regression exceeds the threshold value. Currently, the threshold is set to 1, meaning that the workflow only fails when the new commit is twice as slow as the original one.

Need a better strategy to calculate the performance gap between the original commit and the new commit to generate a desired and accurate threshold value.

Related issue - https://github.com/amazon-ion/ion-java-benchmark-cli/issues/53

cheqianh:
Some thoughts on this.

In the short term, we should increase the warm-up and iteration numbers to make sure the results become stable.

in the long term, 

1. We should compare the new commit not only to the most recent one but also to the average value from the past months, to prevent unnoticed false negatives from occurring.

2. We also need to reduce the noise caused by the machine. Currently, we only run performance regression detection on Ubuntu but we plan to add other OSs in the future. To ensure more consistent benchmark results on Linux, we can control factors such as CPU affinity - https://easyperf.net/blog/2019/08/02/Perf-measurement-environment-on-Linux

3. We should create more specific files that target specific data models. For example, if a new commit introduces changes that impact the speed of string serialization, it can be hard to find. By having files that include numerous test cases for Ion string values, we can magnify this regression and easily detect it.

6. We can also mimic JMH (Java Microbenchmark Harness) and implement another warm-up mode based on 'clock time' instead of 'warm-up execution numbers' which ensures that the benchmark-cli warm up for a specific duration. For example, instead of executing the specific benchmark code 20 times, we warm up the benchmark-cli for 20 minutes.

cheqianh:
Few TODOs:
- We can measure the percentile of the benchmark results - https://learn.microsoft.com/en-us/previous-versions/msp-n-p/bb924370(v=pandp.10)
- We can try running each command for both previous and current commits together and see if it has less noice (currently we are running all commands for each commit each time).

PR #282, opened 2023-08-01
Update benchmark sampling method and statistics

popematt:
**Issue #, if available:**

While not necessarily a fix, this does make significant progress on #265

**Description of changes:**

* Updates/fixes the sampling methodology so that it's properly running samples of 1-2 seconds each.
* Adds a `SampleDist` class that helps compute statistics for a sample distribution
* Drops the "rate" metrics and adds operations per second metrics.
* Drops the percentile stats and adds error and standard deviation.
* Refactors the report generator to make it less spaghetti-like
* Updates the table output format so that it can be directly copy-pasted to github as a markdown table.
* Updates the `compare` command so that the fields argument uses the same field names and syntax as the other commands.
* Drops support for Python 3.7 (which is EOL) because the `statistics.NormalDist` class requires at least Python 3.8.

The perf testing workflow is expected to fail as I have not updated that. It will come in the next PR.


amazon/ionbenchmark/ion_benchmark_cli.py
```diff
@@ -284,7 +294,15 @@ def _run_benchmarks(specs: list, report_fields, output_file):
...
-    print(tabulate(report, tablefmt='fancy_grid', headers='keys'))
+        # TODO: Add some option to dump or otherwise expose the raw sample data. For now, you can
+        # uncomment the following lines to get the raw results as CSV that can be copy/pasted into a spreadsheet.
+        #
+        # printable_key = benchmark_spec.get_name().replace(" ", "").replace(",", "-")
+        # for _x in [printable_key, *result.timings]:
+        #     print(_x, end=",")
+        # print("")
+
+    print(tabulate(report, tablefmt='pipe', headers='keys', floatfmt='.2f'))
```
popematt, line 305: Using the new table format, you can copy/paste directly from the terminal to e.g. a github comment, and this is what you will see:
| name                                                  |   time_mean(ns) |   time_error(ns) |   ops/s_mean |   ops/s_error |
|:------------------------------------------------------|----------------:|-----------------:|-------------:|--------------:|
| (cbor,dumps,cat.cbor)                                 |          621.56 |             6.53 |   1609325.62 |      16351.13 |
| (cbor2,dumps,cat.cbor)                                |         2559.77 |            20.53 |    390726.48 |       3086.54 |
| (json,dumps,cat.json)                                 |         1619.62 |             6.14 |    617453.26 |       2341.06 |
| (ujson,dumps,cat.json)                                |          698.16 |             3.01 |   1432404.24 |       6223.26 |
| (ion_text,dumps,cat.ion)                              |         9420.41 |            84.96 |    106175.15 |        938.70 |
| (ion_binary,dumps,cat.ion)                            |        10463.80 |            59.24 |     95575.76 |        541.19 |
| Ion Binary, pure Python impl                          |       263795.71 |          1908.30 |      3791.34 |         27.22 |
| Ion Text, pure Python impl                            |       107178.88 |          1369.13 |      9334.16 |        115.81 |
| (self_describing_protobuf,dumps,cat.sd_protobuf_data) |         3974.29 |            55.22 |    251742.58 |       3371.29 |
| (protobuf,dumps,cat.protobuf_data)                    |          285.56 |             2.89 |   3502871.19 |      34420.39 |


Note that this suffers from [false precision](https://en.wikipedia.org/wiki/False_precision) (just like JMH reports do). The number formatting needs to be fixed to show a specific number of significant figures—possibly dependent on the error value for the results.

tgregg (approved):
Nice. The `autorange` feature was a good find.

rmarrowstone (approved):

PR #283, opened 2023-08-01
Refactor performance regression workflow

popematt:
**Issue #, if available:**

None

**Description of changes:**

* Splits the workflow into parallel steps by python version _and_ sample data. Within each parallel job, it tests both read and write for the sample data using the baseline and the tip of the PR.
* Adds a step to cache PIP packages to help reduce the setup time in the parallelized jobs. Each parallel job runs in <10 minutes, so we don't need to wait for ages to see the results.
* Moves configurable parameters that we might want to tweak into environment variables at the start of the file.
* Removes the unnecessary `PR-content-check` workflow

This builds on #282 and must be merged after it. The perf workflow will not run successfully until this PR is rebased on top of #282. If you prefer, I can combine this into a single PR, but I thought it might be easier to review if I break them up.

You can see an example of the new workflow [here](https://github.com/popematt/ion-python/actions/runs/5732003064).
There is still some work to be done on getting the right metric for the performance benchmark comparison. I'm beginning to suspect that ops/s might be more stable for slow functions (e.g. running Ion on PyPy or using very large sample data) because it puts the larger number in the denominator, and we might also find that a proper statistical comparison is necessary to allow us to compare two sample means while taking into account the error from _both_ samples.


.github/workflows/performance-regression.yml
```diff
@@ -1,152 +1,140 @@
-name: Performance Regression Detector
+name: Performance Regression
 
 on:
   pull_request:
-    branches: [ master ]
+    paths-ignore:
+      # Workflow will not run if all changes match these paths
+      - '**/*.md'
```
tgregg, line 7: This is the replacement for the PR-content-check, correct? Can we just opt-in `amazon/ion/**` rather than excluding just the markdown files?

.github/workflows/performance-regression.yml
```diff
@@ -1,152 +1,140 @@
...
 
-            python $cli_script write -w $warmups --iterations $iterations -o $runner_dir/benchmarkResults/${testset}_write/new --io-type buffer --format ion_text --format ion_binary --api load_dump $runner_dir/testData/${testset}.10n
-          done
+  detect-regression:
+    name: Check
+    runs-on: ubuntu-latest
+    needs: [generate-test-data, prepopulate-pip-cache]
+    strategy:
+      matrix:
+        python-version: ['3.9', '3.11', 'pypy-3.8', 'pypy-3.10']
```
tgregg, line 70: I'm thinking we should just test against the most recent supported version of both CPython and PyPy.

tgregg (approved):

rmarrowstone (approved):

.github/workflows/performance-regression.yml
popematt: Yes, we can do that.

.github/workflows/performance-regression.yml
popematt: I created #286 for this discussion so that it doesn't have to block this PR.

PR #284, opened 2023-08-09
Improves the deserialization speed by reducing object initialization overhead

cheqianh:
### Description:
Publish the PR for feedback but will keep it in draft mode until we are confident to merge this change. More changes and tests come soon. 

This PR improved the library's performance by 40-50% through:
(1) migrating to new ionPyObjects without inheritance  (+~20%),
(2) switching the way of object initialization from from_value() to constructor directly (+~10%), and
(3) optimizing Ion struct's underlying storage implementation (+~20%)

#### Unit tests

This PR also fixed more than 2k unit test failures due to the changes of IonPyObjects. Most of failures are because of (1) ion_nature inheritance, (2) from_value invocation, and (3) methods signature differences. 

#### Detail:
 
In detail, create an IonPyObject requires a few steps which causes a lot of overhead. The overheads mainly come from (1) Ion_Nature inheritance, (2) from_value() method invocation overhead and (3) methods refactor overhead. In addition, Ion Struct, the Ion multiMap, produces a lot of overhead due to use of high level abstract classes such as OrderedDict, mutableMapping and MutableSequence. In the Ion Python C extension, we always create an empty Ion Struct first, and recursively call its "add_item" API to add new values. Every time inserting a value into the Struct, the code will check if this is the first value occurs for this key. If it's the first value for the key, it will create a new MultiSequence for the key, otherwise it will just insert the value into the list. Creating new mutableValues is extremely expensive when deserializing large Ion files.  

So this PR first removed [Ion_Nature inheritance](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/simple_types.py#L37) and the [reusable help functions](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/simple_types.py#L136-L144) for creating IonPyObject classes.  

Then, all new IonPyObjects has its own constructor. The PR next switched the way of IonPyObjects initialization from `from_value()` to their constructors directly to avoid unnecessary method invocation overhead. 

Last, the PR optimized the [multiMap implementation](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/core.py#L605) by using native list instead of MutableSequence. In addition, since the Ion Struct is an unordered collection, we uses a Python dict directly instead of previous OrderedDict. 

#### Benchmarking
The performance gain depends on various factors such as data model and size. Below are three sample data I used for benchmarking:
|File_name|Description|Size (bytes)|Before (s)|After(s)|Improvement|
|--|--|--|--|--|--|
|committed.10n|a small Ion struct|148|0.19|0.13|31.6%|
|complex.10n|an Ion struct with decimal values|5208|14.1|7.8|46.9%|
|legacy.10n|complex nested Ion containers|22.3e^6|245|127|48%|

#### Next Step
1. There are more experiment and tests required, so we don't break any existing use case
2. All above are focusing on the read(deserialization) process, try write process as well.
3. more IonPyObjects migration for rest types such as timestamp and bytes. 
4. Using the benchmark-cli to generate an official benchmark result.


JustinDaleGray:
**High Level Overview**
- Consider the use of [`__slots__`](https://docs.python.org/3/reference/datamodel.html#slots) versus the use of `__dicts__` or `__weakref__`. We can further reduce memory allocations by not creating a dictionary for attributes where possible. If a dictionary is explicitly required it can be declared within the slot.
- Avoid unnecessary redundancy. Consider wrapping the type classes in module in a base class and moving similar methods. This will mean the use of polymorphism and using some vtable lookups. It might be useful to see what the penalty for DRY might be in this context.
- Consider alternative data structures. For example, `__setitem__(self, key, value)` wraps the value in a list. If only the value is needed, then consider assigning it directly instead.
- Avoid unnecessary conditional checks. Checking a conditional to set a value to `None` may be suboptimal.
- I'm not keen on the `foo_new()` naming convention. Consider alternative naming conventions that adhere to the [PEP 8](https://pep8.org/#class-names) standard on class naming.

rmarrowstone:
At a high-level I think we need to commit to dropping some of the existing factory methods. I know it's arguably part of our API, so we'd need to rev at least minor, but probably major version, but I think it's worth it.

rmarrowstone:
Thanks for doing the work to get to this place! I'm super excited to see the benefits of this! Let's not be afraid to be bold in breaking API compatibility to do so.

cheqianh:
Okay so,
1. the next commit will address above comments. 
2. As for the scope and goal of this PR, I'm going to make break changes and completely deprecate old methods such as `from_value`. This requires a little more work, I can create a separate PR targeting this one as needed.

tests/test_core_multimap.py
```diff
@@ -1,6 +1,7 @@
 from typing import Sequence, NamedTuple
 
-from amazon.ion.core import Multimap
+# from amazon.ion.core import Multimap
+from amazon.ion.ion_py_objects import IonPyDict_new as Multimap  # replace the original multiMap with the new IonPyDict
```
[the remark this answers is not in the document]
cheqianh: Okay. I'm also working on adding more tests.

amazon/ion/equivalence.py
```diff
@@ -19,11 +19,21 @@
...
 
 from amazon.ion.core import IonType, Timestamp, TimestampPrecision, MICROSECOND_PRECISION, OffsetTZInfo, Multimap
+from amazon.ion.ion_py_objects import IonPyNull_new, IonPyDecimal_new, IonPyInt_new, IonPyFloat_new, IonPyText_new, \
+    IonPyList_new, IonPyDict_new, IonPyBool_new, IonPySymbol_new
 from amazon.ion.simple_types import _IonNature, IonPyList, IonPyDict, IonPyTimestamp, IonPyNull, IonPySymbol, \
     IonPyText, IonPyDecimal, IonPyFloat
 from amazon.ion.symbols import SymbolToken
 
 
+def isinstance_of_ion_nature(obj):
```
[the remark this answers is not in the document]
cheqianh: Will fix this in the next commit, or we will just remove all of these things once we deprecate the `from_value` and ion_nature related.

amazon/ion/ioncmodule.c
```diff
@@ -1577,6 +1578,18 @@ PyObject* ionc_init_module(void) {
     _ionpydict_cls              = PyObject_GetAttrString(_simpletypes_module, "IonPyDict");
     _ionpydict_fromvalue        = PyObject_GetAttrString(_ionpydict_cls, "from_value");
 
+    // A temporary workaround to replace from_value() with class constructors
+    // TODO deprecate from_value later
```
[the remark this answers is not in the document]
cheqianh: I left another overall comment below to discuss the scope and goal for this PR.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+    ion_type = IonType.STRUCT
+
+    __name__ = 'IonPyDict_new'
+    __qualname__ = 'IonPyDict_new'
+
+    # one good initialization example: my_dict = IonPyDict(ion_type, value, annotations)
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+        self.__store = {}
+        if args is not None:
```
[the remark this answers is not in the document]
cheqianh: Let me keep this comment open, I'll do some experiment and get back to you later.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+    ion_annotations = ()
+    ion_type = IonType.STRUCT
+
+    __name__ = 'IonPyDict_new'
+    __qualname__ = 'IonPyDict_new'
+
+    # one good initialization example: my_dict = IonPyDict(ion_type, value, annotations)
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+        self.__store = {}
```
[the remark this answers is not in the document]
cheqianh: I will use OrderedDict again. I benchmarked it locally, It will not have a significant negative impact on performance.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+                field_name = SymbolToken(field_name, 0 if field_name is None else None)
+        else:
+            field_name = None
+
+        return IonEvent(event_type, ion_type=self.ion_type, value=value, field_name=field_name,
+                        annotations=self.ion_annotations, depth=depth)
+
+
+class IonPyDecimal_new(Decimal):
+    ion_annotations = ()
```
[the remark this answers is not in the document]
cheqianh: Did you mean that why not define ion_annotation in `__init__`? Since the method signature might change later so I kept both `ion_type` and `ion_annotation` defined at the same place.  

Correct me if this doesn't answer your question but since I saw similar comments above, I can provide my experiment results. This is kind of related to @JustinDaleGray's first feedback.

> Consider the use of [__slots__](https://docs.python.org/3/reference/datamodel.html#slots) versus the use of __dicts__ or __weakref__. We can further reduce memory allocations by not creating a dictionary for attributes where possible.
 
I first used \_\_slots\_\_ to limit the attributes we want, and I saw a 5% of _object initialization_ improvement. 

Later on I found another approach to optimize object initialization speed by setting `class attributes` instead of `instance attributes`. `instance attributes` refers to the attributes defined by \_\_init__. I saw a small _object initialization_ improvement as well.

However, Python doesn't allow both `__slot__` and `class attributes` optimizations at once. I chose the class attributes optimization since we will change the method signature later and all values are default to its own Ion type and empty annotations unless users assign specific values to them.

I just switched between these three approaches and re-run the deserialization benchmarking, I only saw a 1-2% differences between them (\_\_slot__ > class variables > instance variables). So any of them should work. Do you have any concern for class variables?

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+
+    def _copy(self):
+        args, kwargs = self._to_constructor_args(self)
+        value = self.__class__(*args, **kwargs)
+        value.ion_type = self.ion_type
+        value.ion_annotations = self.ion_annotations
+        return value
+
+    @staticmethod
+    def _to_constructor_args(value):
```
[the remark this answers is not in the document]
cheqianh: Yep. This will be replaced by constructor once we fully deprecated the from_value.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+            args, kwargs = cls._to_constructor_args(ion_event.value)
+        else:
+            args, kwargs = (None, None, ()), {}
+        value = cls(*args, **kwargs)
+        value.ion_type = ion_event.ion_type
+        value.ion_annotations = ion_event.annotations
+        return value
+
+    @classmethod
+    def from_value(cls, ion_type, value, annotations=()):
```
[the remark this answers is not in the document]
cheqianh: Yep, I left it here in case of breaking any existing use case and unit test. But as you mentioned above about the break change, I can remove all of them.

Another interesting thing is that I saw the performance improvement by switch between using and not using this `from_value`. It can improve the performance by ~5-10%.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+    __qualname__ = 'IonPyNull_new'
+
+    def __init__(self, ion_type=None, value=None, annotations=()):
+        self.ion_type = ion_type
+        self.ion_annotations = annotations
+
+    def __bool__(self):
+        return False
+
+    def _copy(self):
```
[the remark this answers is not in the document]
cheqianh: Okay, this might be one of the break changes.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
...
+
+
+class IonPyNull_new(object):
+    ion_annotations = ()
+    ion_type = IonType.NULL  # TODO initialized to NULL type first, what's the real type?
+
+    __name__ = 'IonPyNull_new'
+    __qualname__ = 'IonPyNull_new'
+
+    def __init__(self, ion_type=None, value=None, annotations=()):
```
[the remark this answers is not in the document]
cheqianh: In order to not break any existing use case, and easily switch between old and new constructor, I kept the signature of the new `constructor` and `from_value` the same. So both of them are following the same signature template: 
`some_method(ion_type, value, annotation)`

Once we desperate `from_value`, we can change the signature for better customer experience, and we are likely to do this in the PR.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
+from decimal import Decimal
+from collections.abc import MutableMapping
+
+from amazon.ion.core import IonType, IonEvent
+from amazon.ion.symbols import SymbolToken
+
+
+class IonPyNull_new(object):
+    ion_annotations = ()
+    ion_type = IonType.NULL  # TODO initialized to NULL type first, what's the real type?
```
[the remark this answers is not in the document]
cheqianh: Refer to https://github.com/amazon-ion/ion-python/pull/284#discussion_r1290528903

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,605 @@
+from decimal import Decimal
+from collections.abc import MutableMapping
+
+from amazon.ion.core import IonType, IonEvent
+from amazon.ion.symbols import SymbolToken
+
+
+class IonPyNull_new(object):
```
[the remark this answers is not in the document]
cheqianh: Okay, I will keep this comment unsolved until renaming them. 

I named them `object`, `object_new` and `object_original` so that we can separate them in the existing tests for easier debug and performance comparisons.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
rmarrowstone: So let's be clear on our object model and `class` vs `instance` attributes [this SO post](https://stackoverflow.com/questions/12409714/python-class-members) covers the latter well.

In our case we need them to be an instance attribute because each value has it's own annotations.

In the code here you're setting a `class` attribute for `annotations` to the empty tuple. In your `__new__` you are setting an instance attribute that is effectively shadowing the class attribute. So given that you're always shadowing it I don't understand why it exists. 

My concern is that shadowing a class attribute with an instance attributes seems likely to introduce subtle bugs. I would avoid it, particularly since given that you are always shadowing it, I don't understand how the code I see here is more efficient for having it.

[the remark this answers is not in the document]
rmarrowstone: For `ion_type` it makes perfect sense. I wonder if the optimizations you observed were actually due to just the benefit of setting the `ion_type` once per class?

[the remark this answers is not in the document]
cheqianh: Okay will add the change in the coming commit.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
cheqianh: I  left `_copy()` for now and added `__copy__`. We can deprecate `_copy()` when we make breaking changes. (The only use case within the library I found is for [unit tests](https://github.com/amazon-ion/ion-python/blob/master/tests/test_equivalence.py#L336))

tests/test_core_multimap.py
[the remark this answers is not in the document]
cheqianh: I created a separate test file - `test_multimap_IonPyDict` with the same unit tests. The only difference is that the new tests use `IonPyDict` instead of previous `multimap`.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
rmarrowstone: In theory, users shouldn't be using an `_` method.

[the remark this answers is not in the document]
cheqianh: Okay I'll just remove them in the coming commit.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
cheqianh: As discussed offline, I'll remove `args*` and set the default value for the fixed parameter.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
cheqianh: I moved `ion_type` to be a class attribute, for specific type such as List and Sexp, I leave it in the constructor so it won't break any existing use case.

tests/test_core_multimap.py
```diff
@@ -1,7 +1,6 @@
 from typing import Sequence, NamedTuple
```
cheqianh, line 1: Change this test back for the original Multimap, we had a new copy for the new IonPyDict.

amazon/ion/equivalence.py
```diff
@@ -26,7 +26,7 @@
 from amazon.ion.symbols import SymbolToken
 
 
-def isinstance_of_ion_nature(obj):
+def isinstance_of_ion_nature_or_new_ion_py_objects(obj):
```
cheqianh, line 29: This method will be deprecated with other old methods later.

amazon/ion/ion_py_objects.py
```diff
@@ -437,19 +423,19 @@ def to_event(self, event_type, field_name=None, in_struct=False, depth=None):
...
+    ion_type = IonType.LIST
 
-    def __init__(self, ion_type, value, annotations=()):
+    def __init__(self, ion_type=IonType.LIST, value=None, annotations=()):
+        if value is None:
+            value = []
         super().__init__(value)
         self.ion_annotations = annotations
         # it's possible to be Sexp
         self.ion_type = ion_type
```
cheqianh, line 436: the constructor still can control ion_type as it's possible to be `Sexp`.

amazon/ion/equivalence.py
rmarrowstone: Right... But what will replace it? I ask because _I think_ that would be either just trying to get the `iontype` and `annotations` or testing if the two objects `hasattr` them. So why not do  that now and avoid the series of `or`s?

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
rmarrowstone: Please confirm this is still the plan.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+from amazon.ion.symbols import SymbolToken
+
+
+class IonPyNull_new(object):
+    __name__ = 'IonPyNull_new'
+    __qualname__ = 'IonPyNull_new'
+    ion_type = IonType.NULL
+
+    def __init__(self, ion_type=IonType.NULL, value=None, annotations=()):
+        self.ion_type = ion_type  # TODO initialized to NULL type first, what's the real type?
```
rmarrowstone, line 14: what is the action here? I would say it's OK as it is... or suggest that for `IonPyNull` we could avoid the class attribute. But I don't understand if this points to a deeper issue.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+        value.ion_type = ion_type
+        value.ion_annotations = annotations
+        return value
+
+    def to_event(self, event_type, field_name=None, in_struct=False, depth=None):
+        value = self
+        if isinstance(self, IonPyNull_new) or self.ion_type.is_container:
+            value = None
+
+        if in_struct:
```
rmarrowstone, line 58: seems like we could avoid the logic and simplify the parameters by passing only passing `field_name` if it were in a struct. realize that _may_ be a breaking change, although I'd like to understand what behavior is expected when passing a field_name with in_struct=False

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+        else:
+            args, kwargs = cls._to_constructor_args(value)
+            value = cls(*args, **kwargs)
+        value.ion_type = ion_type
+        value.ion_annotations = annotations
+        return value
+
+    def to_event(self, event_type, field_name=None, in_struct=False, depth=None):
+        value = self
+        if isinstance(self, IonPyNull_new) or self.ion_type.is_container:
```
rmarrowstone, line 113: when would either of these be true?

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+            field_name = None
+
+        return IonEvent(event_type, ion_type=self.ion_type, value=value, field_name=field_name,
+                        annotations=self.ion_annotations, depth=depth)
+
+
+class IonPyList_new(list):
+    __name__ = 'IonPyList_new'
+    __qualname__ = 'IonPyList_new'
+    ion_type = IonType.LIST
```
rmarrowstone, line 428: like with null I would not have the class attribute for List.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+
+class IonPyDict_new(MutableMapping):
+    __name__ = 'IonPyDict_new'
+    __qualname__ = 'IonPyDict_new'
+    ion_type = IonType.STRUCT
+
+    def __init__(self, ion_type=IonType.STRUCT, value=None, annotations=()):
+        super().__init__()
+        self.ion_annotations = annotations
+        self.__store = {}
```
rmarrowstone, line 494: Per the resolved comment thread this was to change to `OrderedDict`.

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+        if value is not None:
+            for key, value in iter(value.items()):
+                self.__store[key] = [value]
```
rmarrowstone, line 497: doesn't this overwrite when `value.items()` has multiple mappings for a field?

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+
+    def __getitem__(self, key):
+        return self.__store[key][len(self.__store[key]) - 1]  # Return only one in order not to break clients
+
+    def __delitem__(self, key):
+        del self.__store[key]
```
rmarrowstone, line 503: what is the contract for multiply-mapped fields?

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+    def iteritems(self):
+        for key in self.__store:
+            for value in self.__store[key]:
+                yield key, value
+
+    def items(self):
+        output = []
+        for k, v in self.iteritems():
+            output.append((k, v))
+        return output
```
rmarrowstone, line 539: why does this exist?

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+            if not isinstance(field_name, SymbolToken):
+                field_name = SymbolToken(field_name, 0 if field_name is None else None)
+        else:
+            field_name = None
+
+        return IonEvent(event_type, ion_type=self.ion_type, value=value, field_name=field_name,
+                        annotations=self.ion_annotations, depth=depth)
+
+
+class IonPyDict_new(MutableMapping):
```
rmarrowstone, line 486: If there is existing pydoc about the contract for setting, deleting and accessing fields with multiple mappings please ensure it is moved. If not, now is a great time to add it! :)

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,586 @@
...
+        else:
+            args, kwargs = cls._to_constructor_args(value)
+            value = cls(*args, **kwargs)
+        value.ion_type = ion_type
+        value.ion_annotations = annotations
+        return value
+
+    def to_event(self, event_type, field_name=None, in_struct=False, depth=None):
+        value = self
+        if isinstance(self, IonPyNull_new) or self.ion_type.is_container:
```
rmarrowstone, line 473: here again. please look at all the `to_event` implementations and check why we need this.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
cheqianh: Yep, I'll rename it once we address other comments. I'm using these class names for easier before/after performance comparison...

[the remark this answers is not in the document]
cheqianh: Sorry I missed the updated question. Would you also refer to removing the old IonPyObject? yeah I think we can completely switch to the _new IonPyObject and rename it to the official `IonPyObject`. This won't cause breaking changes.

amazon/ion/equivalence.py
cheqianh: Got it, refactored it to check if both ion_type and ion_annotations attributes exist.

amazon/ion/ion_py_objects.py
cheqianh: Removed.

amazon/ion/ion_py_objects.py
cheqianh: I didn't spend too much effort on the `to_event` API since it's mainly used by the pure python implementation. But I took a look, it seems that we do not have to maintain `field_name`. It requires more investigation to see if this will cause breaking changes (at least some low-level internal methods need this parameter but the top-level dump doesn't).

cheqianh:
A new commit is available below. 

An overall comment: This PR incorporated the performance improvement into the code with the least changes so it won't break any clients use case. Some methods are copied from the original IonPyObject but it's time to refactor/deprecate them! Also Added some documents.

amazon/ion/ion_py_objects.py
cheqianh: `to_event` was same as the original factory methods, which was used by both container and non-container Ion types so it has this condition check. I refactored them.

amazon/ion/ion_py_objects.py
cheqianh: Removed.

amazon/ion/ion_py_objects.py
cheqianh: Thanks for catching this. Reverted by accident when I benchmark before/after commits..

amazon/ion/ion_py_objects.py
cheqianh: The parameter of the original multimap is limited to Python's native dictionary. I added logic to support multi-value Map too.

amazon/ion/ion_py_objects.py
cheqianh: It deletes everything matching the given key currently.

amazon/ion/ion_py_objects.py
cheqianh: Would you refer to `items()` vs.`iteritems()`?

Python 3 has deprecated `iteritems`. I removed `iteritems` in the coming commit too.

amazon/ion/ion_py_objects.py
cheqianh: Sounds good! I'm working on adding a pydoc according to current behavior and we can discuss if we need other ways to distribute method documents.

amazon/ion/ion_py_objects.py
cheqianh: The python 3 `dict.items()` returns a set-like object providing a view on the dictionary's items.

Our `IonPydict.item()` leverage the python generator to iterate the IonPyDict's items.

This Python 2/3 and dict/IonPyDict behavior difference need to be fixed or documented. I'll create a Github issue and see if we still need this method and why.

amazon/ion/ioncmodule.c
```diff
@@ -1577,6 +1578,18 @@ PyObject* ionc_init_module(void) {
...
+    _ionpyint_fromvalue = _ionpyint_cls;
+    _ionpynull_fromvalue = _ionpynull_cls;
+    _ionpyfloat_fromvalue = _ionpyfloat_cls;
+    _ionpydecimal_fromvalue = _ionpydecimal_cls;
+    _ionpytext_fromvalue = _ionpytext_cls;
+    _ionpylist_fromvalue = _ionpylist_cls;
+    _ionpydict_fromvalue = _ionpydict_cls;
+    _ionpybool_fromvalue = _ionpybool_cls;
+    _ionpysymbol_fromvalue = _ionpysymbol_cls;
```
tgregg, line 1591: Should we remove the lines like
```
_ionpynull_fromvalue        = PyObject_GetAttrString(_ionpynull_cls, "from_value");
```
above, which these seem to overwrite?

amazon/ion/ion_py_objects.py
cheqianh: Created GH issue - https://github.com/amazon-ion/ion-python/issues/288

tests/test_simpleion.py
```diff
@@ -376,8 +377,8 @@ def _simple_loads(data, *args, **kw):
 @parametrize(
     *tuple(chain(
         generate_scalars_binary(SIMPLE_SCALARS_MAP_BINARY),
-        generate_containers_binary(_SIMPLE_CONTAINER_MAP),
-        generate_annotated_values_binary(SIMPLE_SCALARS_MAP_BINARY, _SIMPLE_CONTAINER_MAP),
+        # generate_containers_binary(_SIMPLE_CONTAINER_MAP),
+        # generate_annotated_values_binary(SIMPLE_SCALARS_MAP_BINARY, _SIMPLE_CONTAINER_MAP),
```
tgregg, line 381: Should these still be commented out?

amazon/ion/ion_py_objects.py
```diff
@@ -0,0 +1,616 @@
...
+
+from amazon.ion.core import IonType, IonEvent
+from amazon.ion.symbols import SymbolToken
+
+
+class IonPyNull_new(object):
+    __name__ = 'IonPyNull_new'
+    __qualname__ = 'IonPyNull_new'
+
+    def __init__(self, ion_type=IonType.NULL, value=None, annotations=()):
```
tgregg, line 13: Can we drop the `value` parameter? Or is it needed for consistency with the other `__init__` implementations?

amazon/ion/ion_py_objects.py
tgregg: Could we call this IonPySequence or something, then? Or does this just need to conform to the existing naming?

tgregg (approved):
Nice job on these performance improvements! Just minor comments from me.

amazon/ion/ion_py_objects.py
cheqianh: I tried to keep the constructor's signature same as `from_value`, so that anywhere which is using `from_value` can be in-place replaced with their constructor. But it's not necessarily, once the this PR is approved. I will create a separate PR targeting this one for clean up works.

amazon/ion/ion_py_objects.py
cheqianh: Yep, It won't break any current use case since we can call it any name we want (E.g., we are calling it an unofficial name with the `_new` suffix). Since the original list and Sexp are called IonPyList, this might affect some write use case (Not sure, maybe for the read use case, people just need to use IonPyList as a native list so they don't need to worry about its class name?).


Or maybe we can redeclare it to ionPyList in `simple_types.py` first, then fully deprecate it when we make breaking changes in the future. 
```
ionPyList = ionPySequence
```

amazon/ion/ion_py_objects.py
cheqianh: I tried to maintain the constructors' signatures identical to `from_value`'s, so that we can perform an in-place replacement. This approach can also help us quantify the overhead caused by `from_value`. For example, switching from `from_value` to their constructors introduces an 8-12% improvement in performance.

This change is not necessarily after the PR is merged. Once this PR is approved, I will create a separate PR targeting this one for all clean up works.

tests/test_simpleion.py
cheqianh: Changed.

amazon/ion/ioncmodule.c
cheqianh: Yep, I added a new action item 2 to the next step list above. Having this methods overwrite here can switch between `from_value()` and constructor for easier benchmarking.

amazon/ion/ion_py_objects.py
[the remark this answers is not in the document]
cheqianh: The latest commit b6fe2df renames IonPyObject_new to IonPyObject, we can safely remove `IonPyObject_original` now.

rmarrowstone (approved):

cheqianh:
The clean up commit did below:
* Deprecated `ion_nature`
* Migrated objects construction from `from_value()` to their constructors in C extension
* Added IonPytimestamp and IonPyBytes
* Replaced all the original IonPyObject and factory methods with the new IonPyObjects.
* Worked on the IonPyTimestamp signature to make it compatible with the amazon.core.Timestamp
* Updated the unit tests to accommodate the changes.

tgregg (approved):

PR #294, opened 2023-11-02
Adds common examples for the simpleion load/dump APIs.

cheqianh:
Some customers found that the lack of documentation makes them really hard to onboard Ion. While discussing and documenting a long-term solution, I'll use this PR to add common examples for the simpleion load/dump APIs first.


rmarrowstone:
This is a good start.

A couple other observations/suggestions:
1. The main confusion I have as a user is `load` vs `loads` I would clarify that somewhere.
2. To that end: we have a module pydoc. Consider enhancing that to aim users to the primary entry points.
3. After the addition of the c-extension it looks like the bulk of the pydoc is now on the `_python` methods.
4. Personally, I find the type mapping tables (on both loads and dumps) to be too much there and just sort of confusing. A possible suggestion would be to pull that out into the module doc in a single table that had both the Python -> Ion and Ion -> Python type mappings.

rmarrowstone (approved):

tgregg (approved):

PR #295, opened 2023-11-07
Enhance IonPyObjects' constructor signatures

cheqianh:
### Issue Statement:
Currently, IonPyObjects' constructors only support a single type of signature to ensure compatibility with the `from_value()` method so that they can be in-place replaced (E.g., replace `from_value()` with `IonPyObject(...)`) within the C extension. An example of such a signature template is:
```
IonPyFloat(ion_type=IonType.FLOAT, value=56.3, annotations=())
```
Users usually use `from_value()` to construct IonPyObjects. However, some customer usages and tests also employ native python objects initialization signature. For example,
```
IonPyFloat(56.3)
```
which throws an exception
```
TypeError: float() argument must be a string or a number, not 'NoneType'
```
#### Proposed Solution:
So in this PR, I enhance the constructor to allow two kinds of signatures:
(1) from_value like signature
```
IonPyFloat(ion_type=IonType.FLOAT, value=56.3, annotations=())
```
and (2) native python type constructor signature
```
IonPyFloat(value=56.3)
```
If users provide only one parameter, it must be a raw value (E.g., a float value 5.6, an int 12 or a string "string"), and we will use it to construct Ion values. If users provide more than one parameter, we will treat the the signature as the 'from_value'-like method.

#### Performance benchmarking
The change slows down the serialization speed by 6% due to extra arguments length check. 
| name                          |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
|:------------------------------|---------------:|---------------:|----------------:|-----------------------:|
| After |           5208 |      795464.98 |       797990.32 |                 360472 |
| Before |           5208 |      742725.58 |       745995.27 |                 360408 |


tgregg (commented):
Are we changing these from using keyword arguments to positional arguments? Could existing users be constructing these values using the keywords in any order?

rmarrowstone:
Is the driver for this customer feedback or did the recent perf work change how it was working?

cheqianh:
> Are we changing these from using keyword arguments to positional arguments? Could existing users be constructing these values using the keywords in any order?

I tried to make both of them work since both customer use cases exist.

------

While this PR provides a solution to help customers easily transition to version 0.11 and unblock specific use cases, we might want to discuss below alternatives too:

(1) Documenting an official way to construct Ion Value. For example, we can recommend customers to use `IonPyFloat(IonType.FLOAT, 5.6, ())` instead of `IonPyFloat(5.6)`. 

(2) We no longer need to maintain a from_value-like signature. However, this requires lots of refactors and changes in both the C extension and unit tests. 

Currently we have a signature like below, which requires users to provide IonType first:
```
IonPyFLOAT(ion_type=IonType.FLOAT, value=None, annotation=())
```

Ideally,  we can have a signature like below: 
```
IonPyFLOAT(value, ion_type=IonType.FLOAT, annotation=())
```
So that users can use both keyword arguments and positional arguments so that we can eliminate the need for additional condition checks and avoiding the 5% performance slowdown. This requires lots of changes, but I think it would be a good long-term solution and is worth giving a try unless you have other concerns or ideas.

cheqianh:
> Is the driver for this customer feedback or did the recent perf work change how it was working? 

We recommend customers to use the `from_value` method to construct Ion values. However, I recently discovered that some users are also using the constructor to initialize Ion values directly since most IonPyObjects are inherited from native Python objects. This PR tried to add flexibility to the constructor signature to avoid errors.

tgregg (approved):
Approving to unblock existing users, with the acknowledgement that we will continue to investigate the best long-term solution.

rmarrowstone (commented):
So the behavior per the test is:
```
    value_1 = IonPyFloat(5.6)
    value_2 = IonPyFloat(IonType.FLOAT, 5.6, ())
```

I think that's likely to surprise folks. Positional arguments don't typically work that way in python. The way that folks expect something like this to be solved is with keyword arguments.

If what we want is folks to be able to call:
```
 value_1 = IonPyFloat(5.6)
```
(which makes perfect intuitive sense to me!!!)

Then why not simply flip the order of the existing keyword arguments?

cheqianh:
For reference: As per Discussion offline, I'll address [the second long-term solution](https://github.com/amazon-ion/ion-python/pull/295#issuecomment-1802961312) and deprecate the `from_value` API.

PR #298, opened 2023-11-15
Changes IonPyObjects' constructor to be same as their parents' classes.

cheqianh:
### Description:

This PR is working on fixing https://github.com/amazon-ion/ion-python/issues/297. Users use the constructor and `from_value` method to construct Ion values. After the deprecation of IonNature, the constructor of IonPyObjects now have a `from_value`-like signature.

#### Problem statement:
there are a few concerns we have, we need to

In this PR, I allowed both (1) Users should be able to use `from_value` method to construct Ion values.
(2) Users should be able to use IonPyObjects' constructor to construct Ion values, where it's inherited from their parent class (E.g., most of them are native python class such as Int, string)

However, to minimize the risk of breaking existing usage, the C extension return value constructions fall back to using `from_value` method so that IonValues constructions can reuse the existing signature.

### Perf Benchmarking
| name                          |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
|:------------------------------|---------------:|---------------:|----------------:|-----------------------:|
| Before the construction refactor PR |           5208 |     1353876.58 |      1359532.68 |                 475016 |
| Before this PR|           5208 |      935335.43 |       937701.30 |                 365776 |
| After this PR |           5208 |      979675.29 |       989224.61 |                 366144 |

which introduces a 5% perf decrement. (3.5% comparing the original refactor PR)

&nbsp;

### DIFF
--- a/amazon/ion/simple_types.py
+++ b/amazon/ion/simple_types.py
@@ -100,9 +100,9 @@ class IonPyDecimal(Decimal):
     __qualname__ = 'IonPyDecimal'
     ion_type = IonType.DECIMAL
 
-    def __new__(cls, ion_type=IonType.DECIMAL, value=None, annotations=()):
-        v = super().__new__(cls, value)
-        v.ion_annotations = annotations
+    def __new__(cls, *args, **kwargs):
+        v = super().__new__(cls, *args, **kwargs)
+        v.ion_annotations = ()
         return v
 
     def __copy__(self):
@@ -114,7 +114,7 @@ class IonPyDecimal(Decimal):
 
     @staticmethod
     def _to_constructor_args(value):
-        return (None, value,), {}
+        return (value,), {}
 
     @classmethod
     def from_event(cls, ion_event):
@@ -157,10 +157,10 @@ class IonPyBytes(bytes):
     __name__ = 'IonPyBytes'
     __qualname__ = 'IonPyBytes'
 
-    def __new__(cls, ion_type=IonType.BLOB, value=None, annotations=()):
-        v = super().__new__(cls, value)
-        v.ion_annotations = annotations
-        v.ion_type = ion_type
+    def __new__(cls, *args, **kwargs):
+        v = super().__new__(cls, *args, **kwargs)
+        v.ion_annotations = ()
+        # v.ion_type = ion_type
         return v
 
     def __copy__(self):
@@ -172,7 +172,7 @@ class IonPyBytes(bytes):
 
     @staticmethod
     def _to_constructor_args(value):
-        return (None, value,), {}
+        return (value, ), {}
 
     @classmethod
     def from_event(cls, ion_event):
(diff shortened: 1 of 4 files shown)


amazon/ion/simple_types.py
```diff
@@ -594,13 +580,11 @@ class IonPyList(list):
...
-        if value is None:
-            value = []
-        super().__init__(value)
-        self.ion_annotations = annotations
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ion_annotations = ()
         # it's possible to be Sexp
-        self.ion_type = ion_type
+        # self.ion_type = ion_type
```
tgregg, line 587: This line should be removed

tgregg (commented):
This looks good, but would you mind adding a few more tests that demonstrate that constructor invocations that were possible before the migration away from _IonNature are still possible after this change?

cheqianh:
> This looks good, but would you mind adding a few more tests that demonstrate that constructor invocations that were possible before the migration away from _IonNature are still possible after this change?

Sure, I had some tests in the PR, but it was reverted by accident. I'll include it in the next commit.

amazon/ion/simple_types.py
cheqianh: Removed.

tests/test_simpleion.py
```diff
@@ -739,3 +739,28 @@ def test_setting_c_ext_flag():
...
+                               ())),
+    (IonPySymbol(text='symbol', sid=None, location=None), IonPySymbol.from_value(IonType.SYMBOL, 'symbol', ())),
+    (IonPyList([]), IonPyList.from_value(IonType.LIST, [], ())),
+    (IonPyDict({}), IonPyDict.from_value(IonType.STRUCT, {}, ())),
+)
+def test_ion_py_objects_construction(v):
+    """These tests are related to the issue https://github.com/amazon-ion/ion-python/issues/297
+
+    Note that IonPyTimestamp.from_value sets the default precision and fraction, while the original constructor doesn't.
+    So, we compare if they represent the same instance; in other words, we set timestamps_instants_only to True.
```
tgregg, line 764:
```suggestion
    So, we compare if they represent the same instant; in other words, we set timestamps_instants_only to True.
```

tgregg (approved):

tests/test_simpleion.py
```diff
@@ -739,3 +739,28 @@ def test_setting_c_ext_flag():
...
+                               ())),
+    (IonPySymbol(text='symbol', sid=None, location=None), IonPySymbol.from_value(IonType.SYMBOL, 'symbol', ())),
+    (IonPyList([]), IonPyList.from_value(IonType.LIST, [], ())),
+    (IonPyDict({}), IonPyDict.from_value(IonType.STRUCT, {}, ())),
+)
+def test_ion_py_objects_construction(v):
+    """These tests are related to the issue https://github.com/amazon-ion/ion-python/issues/297
+
+    Note that IonPyTimestamp.from_value sets the default precision and fraction, while the original constructor doesn't.
+    So, we compare if they represent the same instant; in other words, we set timestamps_instants_only to True.
```
rmarrowstone, line 764:
```suggestion
    Note that the IonPyTimestamp constructor only sets the attributes for the base datetime type which does not have precision or fractional seconds. IonPyTimestamp.from_value sets these attributes after construction.
    So, we compare if they represent the same instant; in other words, we set timestamps_instants_only to True.
```

rmarrowstone (approved):

PR #302, opened 2023-11-17
Warn on ImportError when importing ionc

rmarrowstone:
Before this change an ImportError other than ModuleNotFound would
cause an exception to propagate on import that prevented users
from using simpleion at all.

I've seen reports of two different users encountering this error
for two different reasons, and am confident this would fix both.

closes https://github.com/amazon-ion/ion-python/issues/300

*Issue #, if available:*

*Description of changes:*


cheqianh (approved):
🚢

PR #309, opened 2023-11-28
Build dict for IonPyDict in ioncmodule

rmarrowstone:
Profiling revealed that a significant portion of time was being taken
by adding items to the IonPyDict. This change optimizes the
construction of IonPyDicts by building the dicts in the ioncmodule.

The impact appears to be an improvement of about 20-25% for parsing
the "service_log" benchmark files.

In implementing the PyDict code in the ioncmodule I learned that
since 3.7 the standard dict type is ordered. The equality of the
standard dict aligns with Ion equality and is preferable. I also
learned that PyDict_SetDefault (in the C API) and dict.setdefault()
(in Python) are a more efficient way to implement get|set logic.

I updated the IonPyDict python class to reflect the above.

The definition of an internal _factory class method optimizes
object construction and that pattern should be applied to the other
classes as well.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -58,7 +58,7 @@ static PyObject* _ionpytext_fromvalue;
 static PyObject* _ionpysymbol_fromvalue;
 static PyObject* _ionpybytes_fromvalue;
 static PyObject* _ionpylist_fromvalue;
-static PyObject* _ionpydict_fromvalue;
+static PyObject* _ionpydict_factory;
 
 static PyObject* _ion_core_module;
 static PyObject* _py_ion_type;
@@ -70,6 +70,7 @@ static PyObject* _ion_symbols_module;
 static PyObject* _py_symboltoken_constructor;
 static PyObject* _exception_module;
 static PyObject* _ion_exception_cls;
+static PyObject* _add_item;
 static decContext dec_context;
 
 typedef struct {
@@ -200,35 +201,6 @@ static PyObject* ion_build_py_string(ION_STRING* string_value) {
     // NOTE: this does a copy, which is good.
     if (!string_value->value) return Py_None;
     return PyUnicode_FromStringAndSize((char*)(string_value->value), string_value->length);
-}
-
-/*
- *  Adds an element to a List or struct
- *
- *  Args:
- *      pyContainer:  A container that the element is added to
- *      element:  The element to be added to the container
- *      in_struct:  if the current state is in a struct
- *      field_name:  The field name of the element if it is inside a struct
- */
-static void ionc_add_to_container(PyObject* pyContainer, PyObject* element, BOOL in_struct, ION_STRING* field_name) {
-    if (in_struct) {
-        PyObject* py_attr = PyUnicode_FromString("add_item");
-        PyObject* py_field_name = ion_build_py_string(field_name);
-        PyObject_CallMethodObjArgs(
-            pyContainer,
-            py_attr,
-            py_field_name,
-            (PyObject*)element,
-            NULL
-        );
-        Py_DECREF(py_attr);
-        Py_DECREF(py_field_name);
-    }
-    else {
-        PyList_Append(pyContainer, (PyObject*)element);
-    }
-    Py_XDECREF(element);
 }
 
 /*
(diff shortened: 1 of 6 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -1037,6 +1009,34 @@ static iERR ionc_read_into_container(hREADER hreader, PyObject* container, BOOL
...
+/*
+ *  Adds an element to a List or struct
+ *
+ *  Args:
+ *      pyContainer:  A container that the element is added to
+ *      element:  The element to be added to the container
+ *      in_struct:  if the current state is in a struct
+ *      field_name:  The field name of the element if it is inside a struct
+ */
+static void ionc_add_to_container(PyObject* pyContainer, PyObject* element, BOOL in_struct, ION_STRING* field_name) {
```
rmarrowstone, line 1021: Slight apologies for moving and changing, but it was super annoying that it was 1000 lines above the one place that called it!

amazon/ion/ioncmodule.c
```diff
@@ -1037,6 +1009,34 @@ static iERR ionc_read_into_container(hREADER hreader, PyObject* container, BOOL
...
+        // this builds the "hash-map of lists" structure that the IonPyDict object
+        // expects for its __store
+        PyObject* py_field_name = ion_build_py_string(field_name);
+        PyObject* empty = PyList_New(0);
+        // SetDefault performs get|set with a single hash of the key
+        PyObject* found = PyDict_SetDefault(pyContainer, py_field_name, empty);
+        PyList_Append(found, element);
+
+        Py_XDECREF(py_field_name);
+        Py_XDECREF(empty);
```
tgregg, line 1032: Was this a conscious decision to change from `Py_XDECREF` from `Py_DECREF`?

amazon/ion/ioncmodule.c
```diff
@@ -1275,19 +1275,27 @@ iERR ionc_read_value(hREADER hreader, ION_TYPE t, PyObject* container, BOOL in_s
...
+
+                    _ionpydict_factory,
+                    data,
+                    py_annotations,
+                    NULL
             );
-            Py_XDECREF(new_dict);
+            if (py_value == NULL) {
+                printf("struct py value is null");
+                PyErr_Print();
```
tgregg, line 1290: Looks like this needs to be cleaned up?

amazon/ion/ioncmodule.c
rmarrowstone: Semi-conscious? I just noted that it is safe to call pointers that may be null and saw it was used already below. I hadn't fully considered the why or why not.

amazon/ion/ioncmodule.c
rmarrowstone: The `printf` and `PyErr_Print`: definitely. But in theory we should _always_ be checking the return values from these calls to the Python runtime and setting/exposing helpful error codes, and we're not anywhere else.

Concretely, if as a developer/tinkerer, I change the name or signature of the python function then this will return NULL and set a PyErr code. Without error checking it's pretty hard to track down what the problem is. I don't necessarily want to add error checking everywhere else, but it would feel odd to have this be the one place we check for NULL and then set an appropriate PyErr to have an exception propagate.

amazon/ion/ioncmodule.c
rmarrowstone: Technically those methods may return NULL pointers, so one would need to use `Py_XDECREF` _if_ we're not checking for NULL returns. Of course if they do return NULL our process is probably cooked anyway.

I'm not super keen to add error checking everywhere so I think using `Py_DECREF` is likely the best choice.

amazon/ion/ioncmodule.c
rmarrowstone: I removed the prints in favor of the existing IERR pattern.

I also created https://github.com/amazon-ion/ion-python/issues/310 to track propagating Python Error information out with the exception that is thrown.

tgregg (approved):

cheqianh (approved):
Thanks for the change!

We now have both the constructor and factory methods. I noticed that they accept different types of parameters. For example, the factory method accepts iterable values. Are we considering promoting it to replace the constructor?

PR #312, opened 2023-12-01
Avoid strcpy of Field Names in C-Ext Read

rmarrowstone:
This change removes a strcpy to a temporary stack allocated array for
the field name of a Struct field. Instead it just allocates the
PyUnicode object for the field name, which does a copy anyway. The
copy is important because the ION_STRING is just a pointer to the
buffer which will change as we descend into parsing the Struct.

The performance impact of this change is measurable for some datasets
but not terribly significant. It is nonetheless an apparent simplifier
and therefore worth doing regardless.

** Please double check my DECREF-ing.**

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -1018,17 +1018,15 @@ static iERR ionc_read_into_container(hREADER hreader, PyObject* container, BOOL 
  *      in_struct:  if the current state is in a struct
  *      field_name:  The field name of the element if it is inside a struct
  */
-static void ionc_add_to_container(PyObject* pyContainer, PyObject* element, BOOL in_struct, ION_STRING* field_name) {
+static void ionc_add_to_container(PyObject* pyContainer, PyObject* element, BOOL in_struct, PyObject* field_name) {
     if (in_struct) {
         // this builds the "hash-map of lists" structure that the IonPyDict object
         // expects for its __store
-        PyObject* py_field_name = ion_build_py_string(field_name);
         PyObject* empty = PyList_New(0);
         // SetDefault performs get|set with a single hash of the key
-        PyObject* found = PyDict_SetDefault(pyContainer, py_field_name, empty);
+        PyObject* found = PyDict_SetDefault(pyContainer, field_name, empty);
         PyList_Append(found, element);
 
-        Py_DECREF(py_field_name);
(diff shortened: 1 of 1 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -1057,22 +1055,11 @@ iERR ionc_read_value(hREADER hreader, ION_TYPE t, PyObject* container, BOOL in_s
...
-
-    char    field_name_value[FIELD_NAME_MAX_LEN];
-    int     field_name_len = 0;
-    BOOL    None_field_name = TRUE;
+    PyObject*   py_field_name = NULL;
 
     if (in_struct) {
         IONCHECK(ion_reader_get_field_name(hreader, &field_name));
-        field_name_len = field_name.length;
-        if (field_name_len > FIELD_NAME_MAX_LEN) {
```
nirosys, line 1068: oh yea, nice to see something like this go away.

amazon/ion/ioncmodule.c
```diff
@@ -1057,22 +1055,11 @@ iERR ionc_read_value(hREADER hreader, ION_TYPE t, PyObject* container, BOOL in_s
...
-            _FAILWITHMSG(IERR_INVALID_ARG,
-                "Filed name overflow, please try again with pure python.");
-        }
-        if (field_name.value != NULL) {
-            None_field_name = FALSE;
-            strcpy(field_name_value, field_name.value);
```
nirosys, line 1074: Surprised this wasn't an issue previously, ion-c documents ION_STRINGs as not being null-terminated.

rmarrowstone: Hah. It apparently was!
```
>>> from amazon.ion import simpleion
>>> simpleion.loads("{'f\0o': bar}")
[1]    18140 segmentation fault  python3
```

Works Now:
```
>>> from amazon.ion import simpleion
>>> simpleion.loads("{'f\0o': bar}")
{'f\x00o': IonPySymbol(text='bar', sid=None, location=None)}
```

rmarrowstone: Created https://github.com/amazon-ion/ion-tests/pull/84 to add a test for this.

tgregg (approved):
The DECREF usage is correct as far as I can tell.

PR #313, opened 2023-12-06
Fix Bugs with emit_bare_values

rmarrowstone:
This change makes it so that the emit_bare_values flag to the
load_extension() method actually works. It does not plumb the
flag up to the user APIs.

Specifically:
* Bare Decimals are correctly emitted instead of tuples
* Values within a non-bare Sexp may now be bare themselves
* None is correctly Increffed when emitted as a bare value

I also patched a leak with ION_DECIMALs where the read decimal
was not being freed.

**PR Note:**
I'd like to see us expand what we emit as std python values, as I think
it will yield solid performance gains and I don't think most folks will 
miss the "ion-ness". But... "most" is not "all" so I think we will need a
set of flags to allow users to specify where they sacrifice fidelity for
speed and convenience.


tgregg (approved):
Looks good to me

PR #314, opened 2023-12-12
Simplify ION_DECIMAL to Python Decimal Conversion

rmarrowstone:
This change simplifies the conversion of an ION_DECIMAL to a Python
Decimal when reading Decimals. Instead of converting the ION_DECIMAL
to a tuple of (sign, (digit, ...), exponent) it just gets the string
representation and uses the corresponding Decimal constructor.

This turns out to be about 20% faster for marshalling "bare value"
decimals of reasonable size (think amounts of money) and uses less
memory.

I also removed an unnecessary extra byte padding for int conversion,
removed a superfluous error check and clarified with comments.

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -204,22 +204,6 @@ static PyObject* ion_build_py_string(ION_STRING* string_value) {
 }
 
 /*
- *  Converts an ion decimal string to a python-decimal-accept string. NOTE: ion spec uses 'd' in a decimal number
- *  while python decimal object accepts 'e'
- *
- *  Args:
- *      dec_str:  A C string representing a decimal number
- *
- */
-static void c_decstr_to_py_decstr(char* dec_str) {
-    for (int i = 0; i < strlen(dec_str); i++) {
-        if (dec_str[i] == 'd' || dec_str[i] == 'D') {
-            dec_str[i] = 'e';
-        }
-    }
-}
-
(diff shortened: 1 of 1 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -543,6 +527,7 @@ iERR ionc_write_value(hWRITER writer, PyObject* obj, PyObject* tuple_as_sexp) {
 
         ION_DECIMAL decimal_value;
         IONCHECK(ion_decimal_from_string(&decimal_value, decimal_c_str, &dec_context));
+        // todo: don't we need to also free decimal_c_str?
```
tgregg, line 530: I don't believe so. `c_string_from_py` calls [PyUnicode_AsUTF8AndSize](https://python.readthedocs.io/en/latest/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize), the documentation for which says "This caches the UTF-8 representation of the string in the Unicode object, and subsequent calls will return a pointer to the same buffer. **The caller is not responsible for deallocating the buffer.**"

tgregg (approved):
Wow, I'm surprised this is that much faster, but we'll take it.

amazon/ion/ioncmodule.c
rmarrowstone: Thanks for reading what was a "come back to this later" note. :) Will check myself and likely remove.

PR #315, opened 2023-12-13
Enhance the benchmark runner to support multiple top level objects use case.

cheqianh:
### Description:
This PR adds support for multiple top level use case, so that we have an apple-to-apple performance comparison for Ion/JSON/CBOR. 

### Note:
- This PR focuses on `io_type=file`, we might consider adding `io_type=buffer` later but we need to figure out how to fairly use JSON and CBOR's load/dump APIs to iterate each of multiple top level object - https://github.com/amazon-ion/ion-python/issues/325.

- When we benchmark read speed for JSON document that includes multiple top level objects, we suppose that the given data is in JSONL. So we have to modify our JSON sample data - [code here](https://github.com/amazon-ion/ion-python/pull/315/files#diff-6d077184aeda689a4e665e62c8cc7c7bb2f8d33073523539f42a3a94d26b55b2R1)

- I'm not sure if protobuf has the same issue for the multiple top-level objects use case. This need to be figure out when we come back to protobuf in the future - https://github.com/amazon-ion/ion-python/issues/326.

- Three files I used for test locally, I upload them for visible but they will not be included when merge this PR into the main branch. They are:
1 [amazon/ionbenchmark/json_newline_vs_list_repro.py](https://github.com/amazon-ion/ion-python/pull/315/files#diff-061791b385cd9f2ffb42daa826b3896f5765ec9d2ff8182a89058d7f24672669)
2 [amazon/ionbenchmark/format_conversion_correct.py](https://github.com/amazon-ion/ion-python/pull/315/files#diff-b70ae6cd92930d856e7b4de45eff0422f5e468ffd7e8decea9af49f659562bef)
3 [amazon/ionbenchmark/format_conversion_inaccurate.py](https://github.com/amazon-ion/ion-python/pull/315/files#diff-402e60208fa4e59d092a876473c7ca3390d4f406c7b5a3293e7fc7851a66ba5c)


amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -114,21 +114,51 @@ def test_fn():
...
-        if _format.format_is_binary(data_format) or _format.format_is_ion(data_format):
+        if _format.format_is_protobuf(data_format):
             def test_fn():
                 with tempfile.TemporaryFile(mode="wb") as f:
                     return loader_dumper.dump(data_obj, f)
+        elif _format.format_is_binary(data_format) or _format.format_is_ion(data_format):
+            def test_fn():
+                with tempfile.TemporaryFile(mode="ab") as f:
+                    for top_level_obj in data_obj:
+                        loader_dumper.dump(top_level_obj, f)
```
tgregg, line 156: For Ion, is this going to write a new symbol table for each top-level value? That would not lead to the most efficient encoding. How does the performance and data size for Ion compare when using this technique vs using a single dump of the entire list of top-level values?

amazon/ionbenchmark/benchmark_spec.py
```diff
@@ -176,12 +176,37 @@ def get_input_file_size(self):
...
+                    cbor_objects = []
+                    while True:
+                        try:
+                            cbor_objects.append(loader.load(f))
+                        except EOFError:
+                            break
+                    self._data_object = cbor_objects
+            elif format_is_protobuf(format_option):
+                with open(self.get_input_file(), "rb") as fp:
+                    self._data_object = loader.load(fp)
```
tgregg, line 209: If we're not sure whether the protobuf logic is correct, then I'd rather throw an error for now.

amazon/ionbenchmark/format_conversion_inaccurate.py
```diff
@@ -0,0 +1,42 @@
...
+cbor_file_list = 'test_cbor'
+
+# Flags to control whether a file in a specific format needs to be generated.
+convert_json = True
+convert_cbor = True
+
+# Flag to control if the JSON file is delimited by a newline.
+generate_jsonl = True
+
+with open(ion_file, 'br') as fp:
```
tgregg, line 21: For posterity, can you reply here to explain what was wrong with this technique?

tests/test_benchmark_cli.py
```diff
@@ -218,3 +219,16 @@ def test_format_conversion_ion_text_to_ion_binary():
...
+@parametrize(
+    ('write', 'json', generate_test_path('./sample_spec/multiple_top_level_object.json')),
+    ('write', 'cbor2', generate_test_path('./sample_spec/multiple_top_level_object.cbor')),
+    ('read', 'json', generate_test_path('./sample_spec/multiple_top_level_object.json')),
+    ('read', 'cbor2', generate_test_path('./sample_spec/multiple_top_level_object.cbor')),
+)
+def test_multiple_top_level_values(args):
+    (command, format_option, file) = args
+    (error_code, _, _) = run_cli([f'{command}', file, '--format', f'{format_option}', '--io-type', 'file'])
+    assert not error_code
```
tgregg, line 233: Are we able to assert that multiple top-level values are actually read/written?

tests/test_benchmark_spec.py
```diff
@@ -11,6 +14,51 @@ def _generate_test_path(p):
...
+    data_obj = _multiple_top_level_json_obj_spec.get_data_object()
+    obj_count = len(data_obj)
+    load_count = 0
+    with open(join(_generate_test_path("sample_spec"), 'multiple_top_level_object.json'), 'r') as f:
+        # iterate each top level object
+        while True:
+            jsonl = f.readline()
+            if jsonl == '':
+                break
+            # make sure the json object are equivalence
```
tgregg, line 37:
```suggestion
            # make sure the json object are equivalent
```

tests/test_benchmark_spec.py
```diff
@@ -11,6 +14,51 @@ def _generate_test_path(p):
...
+def test_write_generate_multiple_top_level_cbor_values():
+    data_obj = _multiple_top_level_cbor_obj_spec.get_data_object()
+    obj_count = len(data_obj)
+    load_count = 0
+    with open(join(_generate_test_path("sample_spec"), 'multiple_top_level_object2.cbor'), 'br') as f:
+        # iterate each top level object
+        while True:
+            try:
+                o = cbor2.load(f)
+                # make sure the CBOR object are equivalence
```
tgregg, line 54:
```suggestion
                # make sure the CBOR object are equivalent
```

cheqianh:
File `multiple_top_level_object.json` and `multiple_top_level_object.cbor` includes equivalent data:
```
{"name":"John", "age":30, "car":null}
{"name":"Mike", "age":33, "car":null}
{"name":"Jack", "age":24, "car":null}
```

amazon/ionbenchmark/benchmark_runner.py
cheqianh:
```
loader_dumper.dump(data_obj, f, sequence_as_stream=True)
| (ion_binary,dumps,service_log_legacy) |       22263038 |  2027258750.00 |   2035399545.00 |               42571303 |
```
```
loader_dumper.dump(data_obj, f, sequence_as_stream=False)
| (ion_binary,dumps,service_log_legacy) |       22263038 |  2010238558.40 |   2024798020.82 |               42571457 |
```
```
for top_level_obj in data_obj:
    loader_dumper.dump(top_level_obj, f)
| (ion_binary,dumps,service_log_legacy) |       22263038 |  2486109233.40 |   2495525362.50 |                  44581 |
```

Changed to `loader_dumper.dump(data_obj, f, sequence_as_stream=True)`

amazon/ionbenchmark/benchmark_spec.py
cheqianh: Changed.

tests/test_benchmark_spec.py
cheqianh: Changed.

tests/test_benchmark_spec.py
cheqianh: Changed.

amazon/ionbenchmark/format_conversion_inaccurate.py
cheqianh: Sure, the reasons are:
1. Since JSON doesn't support multiple top-level objects, we used to store all JSON object within a list. Now we iterate each top level values and write them one by one. 
2. CBOR conversion was wrong. We used jsonable bytes for JSON/CBOR conversion instead of Python objects that are converted from JSON objects so that the resulted CBOR data is incorrect.

tests/test_benchmark_cli.py
cheqianh: As discussed offline, I'll create a few new tests for each generated test_fun to ensure they work correctly.

amazon/ionbenchmark/benchmark_spec.py
```diff
@@ -176,12 +176,37 @@ def get_input_file_size(self):
...
     def get_data_object(self):
         """
-        Get the data object to be used for testing. Used for benchmarks that write data.
+        Get a list that holds all data objects to be used for testing. Used for benchmarks that write data.
         """
         if not self._data_object:
             loader = self.get_loader_dumper()
-            with open(self.get_input_file(), "rb") as fp:
-                self._data_object = loader.load(fp)
+            format_option = self.get_format()
```
rmarrowstone, line 183: Why have all this branching here? we already have `loader`s for each type. Put the loader specific code there.

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -114,21 +114,51 @@ def test_fn():
         data_file = benchmark_spec.get_input_file()
 
         def test_fn():
-            with open(data_file, "rb") as f:
-                return loader_dumper.load(f)
+            format_option = benchmark_spec.get_format()
+            if _format.format_is_ion(format_option):
```
rmarrowstone, line 118: Put the loader specific logic in custom loaders for each type, avoid branching here.

amazon/ionbenchmark/benchmark_spec.py
cheqianh: Okay, I'll refactor them into a specific file there. I need to run the benchmark-cli again to ensure that the write/read process won't affect the performance (e.g., values generated by yield instead of returning an iterator directly).

cheqianh: Refactor Json and Cbor, the performance are almost the same but the memory usage peak have a signaficant difference, I'm going to see why. 

Also since we dump the whole document for Ion to avoid unnecessary symbol table write, the Ion_load_dump needs to be changed too.

cheqianh:
The commit [74d89ad](https://github.com/amazon-ion/ion-python/pull/315/commits/74d89ad7630434a10dad8dc50071ca9bb9713167) refactor loader (read APIs) to separate files. And below are the new results for a log
| name                                 |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
| ------------------------------------- | --------------- | --------------- | ---------------- | ----------------------- |
| Ion - after                          |       22263038 |  5092575416.60 |   5109052881 |                 276500 |
| Ion - before                         |       22263038 |                |   4730782811    |                 171700 |
||
| json - after |      145330385 |   871989650.00 |    872934369 |                 112139 |
| json - before |      145330385 |   |    875,974,056 |                 77,446 |
||
| cbor - after |      116960762 |   985745625.00 |    991288347 |                  87769 |
| cbor - before |      116960762 |   |    995,595,161 |                  51,222 |

cheqianh:
Below is the table showing the differences in write performance after the refactor.

| name                                 |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
| ------------------------------------- | --------------- | --------------- | ---------------- | ----------------------- |
| Ion - before                         |       22263038 |               |   2,509,895,173    |                 42,573,352 |
| Ion - after                          |       22263038 |  |   4,139,845,225 |               42,573,352  |
||
| json - before |      145330385 |   |    3,981,317,292 |                 143,755,789 |
| json - after |      145330385 |   |   4,018,459,501 |              143,756,301   |
||
| cbor - before |      116960762 |   |    1,303,162,089 |                  19,301 |
| cbor - after |      116960762 |   |   1,291,955,078 |                 19,813 |

tgregg:
Why is Ion read and write performance worse after the refactor?

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -112,23 +116,90 @@ def test_fn():
...
+                    with open(data_file, 'br') as f:
+                        for v in loader_dumper.load(f):
+                            pass
+        elif _format.format_is_protobuf(format_option):
+            # Refer to https://github.com/amazon-ion/ion-python/issues/326
+            raise NotImplementedError("Benchmarking Protocol Buffer multiple top level object use case may not "
+                                      "support yet.")
 
     elif match_arg == ['file', 'write', 'load_dump']:
+        # This method should return a list that holds all top_level values.
```
rmarrowstone, line 169: Why? Why not return a generator?

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -112,23 +116,90 @@ def test_fn():
...
+                        loader_dumper.dump(data_obj, f)
+            else:
+                def test_fn():
+                    with tempfile.TemporaryFile(mode="ab") as f:
+                        loader_dumper.dump(data_obj, f)
+        elif _format.format_is_ion(data_format):
+            if custom_file:
+                def test_fn():
+                    with open(custom_file, 'ab') as f:
+                        loader_dumper.dump(data_obj, f, sequence_as_stream=True)
```
rmarrowstone, line 189: The purpose of the `loader_dumper` abstraction is to encapsulate format stuff like `sequence_as_stream`. Why does that need to be passed here?

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -112,23 +116,90 @@ def test_fn():
...
         data_format = benchmark_spec.get_format()
-        if _format.format_is_binary(data_format) or _format.format_is_ion(data_format):
-            def test_fn():
-                with tempfile.TemporaryFile(mode="wb") as f:
-                    return loader_dumper.dump(data_obj, f)
+        if _format.format_is_protobuf(data_format):
+            # Refer to https://github.com/amazon-ion/ion-python/issues/326
+            raise NotImplementedError("Benchmarking Protocol Buffer multiple top level object use case may not "
+                                      "support yet.")
+        elif _format.format_is_cbor(data_format):
```
rmarrowstone, line 176: There's a lot of duplicated code here. There are three things that vary:
1. custom file or not.
2. binary or text.
3. ion specific flag.
1 and 2 should be pretty easily dealt with by defining what to open before opening it. 3 should be dealt with by putting ion specific stuff in the ion loader dumper.

amazon/ionbenchmark/benchmark_spec.py
rmarrowstone: I wonder if it is due to building the list?

rmarrowstone: I'm also not sure why IonLoadDump needs to change. My understanding is that we're not flushing the writer so it will just buffer until it closes, then flush the symbol table and values. Right?

amazon/ionbenchmark/benchmark_runner.py
cheqianh: Okay yeah, we can reuse the `loader.load` for that. Working on this.

cheqianh:
Re, 
> I'm also not sure why IonLoadDump needs to change. My understanding is that we're not flushing the writer so it will just buffer until it closes, then flush the symbol table and values. Right?

I saw a 20% difference between `dumps` each top-level object and `dump` the whole document. I can take a look at that later.

amazon/ionbenchmark/benchmark_spec.py
cheqianh: Re:
> I'm also not sure why IonLoadDump needs to change. My understanding is that we're not flushing the writer so it will just buffer until it closes, then flush the symbol table and values. Right?

I saw a 20% difference between dumps each top-level object and dump the whole document. I'll try to see why it takes shorter when repeatedly calling dump

amazon/ionbenchmark/benchmark_runner.py
cheqianh: Got it. I had all branches separated for debugging reasons to allow me to navigate the specific formats affected by the refactor. I will include the refactor for these three types in the next commit.

amazon/ionbenchmark/benchmark_runner.py
cheqianh: Changed

cheqianh:
The commit [2b177af](https://github.com/amazon-ion/ion-python/pull/315/commits/2b177af8b85f9a82e49de9aaba9ccc19e79cdb35) addressed the comment. A few comments and highlights below. 
1. Re: https://github.com/amazon-ion/ion-python/pull/315#discussion_r1442404936
It seems that the generator can't be fully consumed by test_fun; it will only parse the first value and return when I pass a generator to it. I refactored the `benchmark_spec.get_data_object()` method to return a generator and reused it in other places to avoid repeating code. However, for benchmarking, we still call `list(the_generator)` to convert it into a list before passing it to `test_fun`. Do you have any concerns or comments about this?

2. Re: https://github.com/amazon-ion/ion-python/pull/315#discussion_r1442408471
I'm going to investigate why this happens and determine which benchmarking methods are more accurate. Based on the previous results, there is a 20% difference between these two approaches.

3. After the refactor, the Ion write performance dropped from 2e^9 to 4e^9 I'm not sure why; I'm investigating it now.

4. There are some conflicts with the main branch. (I guess it's because of the recent bare_value PR) I'll work on them at the end.

rmarrowstone:
>    It seems that the generator can't be fully consumed by test_fun; it will only parse the first value and return when I pass a generator to a defined `test_fun`. I refactored the `benchmark_spec.get_data_object()` method to return a generator and reused it in other places to avoid repeating code. However, for benchmarking, we still call `list(the_generator)` to convert it into a list before passing it to `test_fun`. Do you have any concerns or comments about this?

Can't we just pull off the generator and do nothing with the values in the `test_fun`?

cheqianh:
The highlight No.3 in my [comment](https://github.com/amazon-ion/ion-python/pull/315#issuecomment-1881885164) above is solved and I updated the metrics table for dump APIs.

This is because for the `before` metrics, I checked out the latest ion-python commit to see how much improvement we have made for the write side. However, since the PR is not merged yet so that the `after` result is based on the latest released version 0.11.3. I updated it. In the future I will use this commit [d425587](https://github.com/amazon-ion/ion-python/commit/d425587d94f8e9fa22368ed1b82a1b65fcb81820) for benchmarking comparison in this PR (the current latest commit).

cheqianh:
Investigating the build failure issue, it worked well locally.

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -84,19 +84,23 @@ def run_benchmark(benchmark_spec: BenchmarkSpec):
...
 
     return BenchmarkResult(nanos_per_op, ops_per_sec, peak_memory_usage)
 
 
-def _create_test_fun(benchmark_spec: BenchmarkSpec):
+def _create_test_fun(benchmark_spec: BenchmarkSpec, return_obj=False, custom_file=False):
     """
     Create a benchmark function for the given `benchmark_spec`.
+
+    :param return_obj: If the test_fun returns the load object for debugging. It only works for `io-type=file` and
```
rmarrowstone, line 97: I don't think we're using this pydoc style anywhere else. simpleion uses the google pydoc style. That is the main API to this code, follow what that does.

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -105,30 +109,64 @@ def test_fn():
             return loader_dumper.loads(buffer)
 
     elif match_arg == ['buffer', 'write', 'load_dump']:
+        # This method returns a list
```
rmarrowstone, line 112: remove?

amazon/ionbenchmark/benchmark_spec.py
```diff
@@ -177,12 +181,23 @@ def get_input_file_size(self):
...
+            if format_is_protobuf(format_option):
+                # Refer to https://github.com/amazon-ion/ion-python/issues/326
+                raise NotImplementedError("Benchmarking Protocol Buffer multiple top level object use case may not "
+                                          "support yet.")
+            else:
+                rtn = []
+                with open(read_file, 'br' if format_is_bytes(format_option) else 'r') as fp:
+                    obj = loader.load(fp)
+                    for v in obj:
+                        rtn.append(v)
```
rmarrowstone, line 199:
```suggestion
                    rtn = [v for v in loader.load(fp)]
```

Shorter, and I believe, faster.

amazon/ionbenchmark/ion_load_dump.py
```diff
@@ -26,12 +26,17 @@ def loads(self, s):
...
+        while True:
+            try:
+                yield next(it)
+            except StopIteration:
+                return
 
     def dumps(self, obj):
         ion.c_ext = self._c_ext
-        return ion.dumps(obj, binary=self._binary)
+        ion.dumps(obj, binary=self._binary)
```
tgregg, line 38: Should this use `sequence_as_stream=True`?

amazon/ionbenchmark/cbor_load_dump.py
```diff
@@ -0,0 +1,27 @@
+import cbor2
+
+
+class CborLoadDump:
```
rmarrowstone, line 4: `Cbor2LoadDump`?

amazon/ionbenchmark/ion_load_dump.py
rmarrowstone, line 38: does this need to change too?

amazon/ionbenchmark/json_load_dump.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def load(self, fp):
+        while True:
+            line = fp.readline()
+            if line == '':
+                return
+            yield json.loads(line)
+
+    def loads(self, s):
+        return json.loads(s)
```
rmarrowstone, line 16: so loads can only handle single values?

rmarrowstone (approved):

amazon/ionbenchmark/json_load_dump.py
cheqianh: `io_type=buffer` currently doesn't handle multiple top-level use case, I opened an issue for that - https://github.com/amazon-ion/ion-python/issues/325

amazon/ionbenchmark/ion_load_dump.py
cheqianh: Added back.

amazon/ionbenchmark/cbor_load_dump.py
cheqianh: changed.

amazon/ionbenchmark/benchmark_spec.py
cheqianh: Changed, thanks for catching this up.

amazon/ionbenchmark/benchmark_runner.py
cheqianh: Removed.

amazon/ionbenchmark/benchmark_runner.py
cheqianh: Changed.

amazon/ionbenchmark/Format.py
```diff
@@ -25,11 +25,21 @@ def format_is_cbor(format_option):
...
+def format_is_protobuf(format_option):
+    return (format_option == Format.SD_PROTOBUF.value) or (format_option == Format.PROTOBUF.value)
+
+
 def format_is_binary(format_option):
     return format_is_cbor(format_option) or (format_option == Format.ION_BINARY.value) \
            or (format_option == Format.PROTOBUF.value) or (format_option == Format.SD_PROTOBUF.value)
 
 
+def format_is_bytes(format_option):
```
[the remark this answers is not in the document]
cheqianh: `ion_text` is not a binary format, but it needs a 'b' flag to open the file.

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -105,30 +109,64 @@ def test_fn():
...
+            # Refer to https://github.com/amazon-ion/ion-python/issues/326
+            raise NotImplementedError("Benchmarking Protocol Buffer multiple top level object use case may not "
+                                      "support yet.")
+        elif _format.format_is_cbor(data_format) \
+                or _format.format_is_json(data_format) \
+                or _format.format_is_ion(data_format):
+            if custom_file:
+                if _format.format_is_bytes(data_format):
+                    def test_fn():
+                        with open(custom_file, 'ab') as f:
```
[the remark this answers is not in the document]
cheqianh: Changed. Yeah it works, thanks for the recommendation.

amazon/ionbenchmark/benchmark_runner.py
```diff
@@ -105,30 +109,64 @@ def test_fn():
...
+                        returned_obj.append(value)
+                return returned_obj
+
+    elif match_arg == ['file', 'write', 'load_dump']:
+        # This method returns a list that holds all top_level values.
+        data_obj = benchmark_spec.get_data_object()
+        data_obj = list(data_obj)
 
+        data_format = benchmark_spec.get_format()
+        if _format.format_is_protobuf(data_format):
```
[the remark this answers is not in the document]
cheqianh: I'm not sure if protobuf can or should handle multiple top-level objects. I opened an issue for this - https://github.com/amazon-ion/ion-python/issues/326 and throw an error for now JIC.

amazon/ionbenchmark/ion_load_dump.py
cheqianh: io_type=buffer currently doesn't handle multiple top-level use case, I opened an issue for that - https://github.com/amazon-ion/ion-python/issues/325

amazon/ionbenchmark/Format.py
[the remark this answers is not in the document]
rmarrowstone: I understand the logic requirement. What I don't understand is what 'is_binary' is actually called for other than determining the flag.

amazon/ionbenchmark/ion_load_dump.py
rmarrowstone: The question was as to whether this needed `sequence_as_stream=True`

amazon/ionbenchmark/json_load_dump.py
rmarrowstone: Ok. At first blush it seems like just making multiple TLVs there would be simpler than creating an issue and having divergent code paths, but we need to get this merged.

PR #317, opened 2023-12-14
Reorder simpleion and Refactor Pydoc

rmarrowstone:
This change puts the user API methods at the top of the file
and moves the pydoc to the actual API methods. I followed the std
json module in the choice to put the main doc strings on dump and
load and have dumps and loads refer to them.

It also factors out the tables that were in loads and dumps into
a single table. I found the previous factoring very difficult to grok
and I believe this is more readable.

I also removed all of the unused kwarg parameters and their pydoc
strings. It added a lot of cognitive load for no value.

Finally I made some minor cleanings to Arg descs and corrected the
default text buffer value for loading with the C-Extension.

**Note to Reviewers:**
I've found the organization of the file, the conversion tables and the
unused args confusing for some time. As I looked to expose the
`emit_bare_values` flag I wanted a good pydoc foundation to build
from.

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -12,13 +12,53 @@
 # specific language governing permissions and limitations under the
 # License.
 
-"""Provides a ``simplejson``-like API for dumping and loading Ion data."""
+"""Provides a ``simplejson``-like API for dumping and loading Ion data.
+
+The below table describes how types from the Ion data model map to the IonPy types
+in simple_types.py as well as what other Python types are supported on dump.
+
+        +-------------------+-------------------+-----------------------------------+
+        | Ion Data Type     | IonPy Type        | Other Dump Mappings               |
+        |-------------------+-------------------|-----------------------------------|
+        | null.<type>       | IonPyNull(<type>) | None                              |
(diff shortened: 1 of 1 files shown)


amazon/ion/simpleion.py
```diff
@@ -50,112 +90,142 @@
...
+    """Deserialize Ion values from ``fp``, a file-handle to an Ion stream, as Python object(s) using the
+    conversion table described in the pydoc. Common examples are below, please refer to the
+    [Ion Cookbook](https://amazon-ion.github.io/ion-docs/guides/cookbook.html) for detailed information.
+
+    Read an Ion value from file_handle:
+        ``simpleion.load(file_handle)``
+
+    Read Ion values using an iterator:
+        ``
+        it = simpleion.load(file_handle, parse_eagerly=True)
```
tgregg, line 168:
```suggestion
        it = simpleion.load(file_handle, parse_eagerly=False)
```
Right?

tgregg (approved):

amazon/ion/simpleion.py
rmarrowstone: Good catch. It was like that before I moved it, but I should've caught that myself. Thanks!

PR #319, opened 2023-12-16
Remove PyObject_HasAttrString condition check.

linlin-s:
*Issue #, if available:*
N/A

*Description of changes:*
In this PR, we have eliminated the `PyObject_HasAttrString` condition check because the `PyObject_GetAttrString` function  itself checks whether the attribute exists and will return NULL if it doesn't. If the `PyObject_GetAttrString` returns `NULL`, this indicates that an error has occurred, and we will handle the `AttributeError` under this condition.

Here is the benchmark results from benchmarking ion-python (C extension enabled) using shorthand log data.
There is 21.23% performance improvement (`time_mean`) on shorthand log data and 3.35% performance improvement(`time_mean`) on single nested test data.

**`ion-python-benchmark-cli` usage pattern to reproduce the benchmark results:**
` python ion_benchmark_cli.py write --iterations 30 --warmups 10 --io-type file test_data.ion --format ion_binary`

**shorthand log**
|             |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
| Before |       21270622 |  3739596533.40 |   3798220835.84 |               42668611 |
| After |       21270622 |  2950278533.40 |   2991548144.17 |               42616337 |

**single nested**
|              |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
|:-----------------------------|---------------:|---------------:|----------------:|-----------------------:|
| Before |            795 |      197943.52 |       204293.10 |                  35887 |
| After |            795 |      193288.23 |       197446.85 |                  29246 |

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -147,10 +147,11 @@ static int offset_seconds(PyObject* timedelta) {
  */
 static int ion_type_from_py(PyObject* obj) {
     PyObject* ion_type = NULL;
-    if (PyObject_HasAttrString(obj, "ion_type")) {
-        ion_type = PyObject_GetAttrString(obj, "ion_type");
+    ion_type = PyObject_GetAttrString(obj, "ion_type");
+    if (ion_type == NULL) {
+        PyErr_Clear();
+        return tid_none_INT;
     }
-    if (ion_type == NULL) return tid_none_INT;
     int c_type = c_ion_type_table[PyLong_AsSsize_t(ion_type)];
     Py_DECREF(ion_type);
     return c_type;
@@ -288,11 +289,12 @@ static iERR ionc_write_symboltoken(hWRITER writer, PyObject* symboltoken, BOOL i
 static iERR ionc_write_annotations(hWRITER writer, PyObject* obj) {
     iENTER;
     PyObject* annotations = NULL;
-    if (PyObject_HasAttrString(obj, "ion_annotations")) {
-        annotations = PyObject_GetAttrString(obj, "ion_annotations");
+    annotations = PyObject_GetAttrString(obj, "ion_annotations");
+    if (annotations == NULL || PyObject_Not(annotations)) {
+        PyErr_Clear();
+        // Proceed as if the attribute is not there.
+        SUCCEED();
     }
-
-    if (annotations == NULL || PyObject_Not(annotations)) SUCCEED();
 
     annotations = PySequence_Fast(annotations, "expected sequence");
     Py_ssize_t len = PySequence_Size(annotations);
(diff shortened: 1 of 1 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -557,16 +559,20 @@ iERR ionc_write_value(hWRITER writer, PyObject* obj, PyObject* tuple_as_sexp) {
...
 
         ION_TIMESTAMP timestamp_value;
-        PyObject *fractional_seconds, *fractional_decimal_tuple, *py_exponent, *py_digits;
+        PyObject *fractional_seconds, *fractional_decimal_tuple, *py_exponent, *py_digits, *precision_attr;
         int year, month, day, hour, minute, second;
         short precision, fractional_precision;
         int final_fractional_precision, final_fractional_seconds;
-        if (PyObject_HasAttrString(obj, "precision") && PyObject_GetAttrString(obj, "precision") != Py_None) {
+        precision_attr = PyObject_GetAttrString(obj, "precision");
+        if (precision_attr == NULL) {
```
rmarrowstone, line 567: Do we have a test that tries to write a std `datetime`? I may misunderstand which branches that would hit or the indentation, but I think that would fail with this change.

linlin-s: > Do we have a test that tries to write a std `datetime`? I may misunderstand which branches that would hit or the indentation, but I think that would fail with this change.

This change did fail the build process. I'll look into the issue and fix the failure. We have the test for timestamp(test_timestamp.py) which has passed locally. I might need to look into the test file to see if we have the test for writing std `datetime` and figure out the reason why it's failed in CI/CD workflow.

linlin-s:
Here are the benchmark results after commit [4f4788d](https://github.com/amazon-ion/ion-python/pull/319/commits/4f4788d4ccde0a6956233793129826ffb9cfb5d8)
There is **_20.62%_** performance improvement (time_mean) on shorthand log data and _**3.48%**_ performance improvement(time_mean) on single nested test data.

**shorthand log**
| Before |       21270622 |  3745274325.00 |   3781944997.77 |               42696241 |
| After |       21270622 |  2960032208.40 |   3001857282.51 |                42644049 |

**single nested**
| name                         |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
|:-----------------------------|---------------:|---------------:|----------------:|-----------------------:|
| Before |            795 |      203792.17 |       209076.32 |                  35595 |
| After |            795 |      193143.60 |       201790.09 |                  29303 |

amazon/ion/ioncmodule.c
linlin-s: There is test in [test_roundtrip](https://github.com/amazon-ion/ion-python/blob/c9907633d973918cd1c26a2b710a12ce67db23ba/tests/test_simpleion.py#L596), one of the parameters is [datetime](https://github.com/amazon-ion/ion-python/blob/c9907633d973918cd1c26a2b710a12ce67db23ba/tests/test_simpleion.py#L517). This test was failed because of this change, after updating in commit [4f4788d](https://github.com/amazon-ion/ion-python/pull/319/commits/4f4788d4ccde0a6956233793129826ffb9cfb5d8), this test is passed.

rmarrowstone (approved):

PR #321, opened 2023-12-20
Enhance ionc_write_big_int method for optimized handling of large integers.

linlin-s:
*Issue #, if available:*

N/A

*Description of changes:*

The updated method `ionc_write_big_int` initially attempts to convert the Python object to a C` long long` using `PyLong_AsLongLongAndOverflow`.  If the integer fits within a 64-bit range and no error occurs, it proceeds to write the integer directly using `ion_writer_write_int64`. For integers larger than 64-bit, it will fallback to string conversion. This PR also replaced method `PyObject_CallMethod(obj, "__str__", NULL)` with `PyObject_Str(obj)`, this also improve the performance. 

Here are the benchmark results from benchmarking ion-python writing using service_log_legacy: There is **_14.74%_** performance improvement from the change.

| Before |       21270622 |  2943214116.60 |   3002532105.01 |               42637008 |
| After |       21270622 |  2520604525.00 |   2559838239.71 |               42647684 |

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -420,14 +420,23 @@ fail:
  */
 static iERR ionc_write_big_int(hWRITER writer, PyObject *obj) {
     iENTER;
-    PyObject* int_str = PyObject_CallMethod(obj, "__str__", NULL);
-    ION_STRING string_value;
-    ion_string_from_py(int_str, &string_value);
-    ION_INT ion_int_value;
+    PyObject* int_str = NULL;
+    int overflow;
+    long long int_value = PyLong_AsLongLongAndOverflow(obj, &overflow);
 
-    IONCHECK(ion_int_init(&ion_int_value, NULL));
-    IONCHECK(ion_int_from_string(&ion_int_value, &string_value));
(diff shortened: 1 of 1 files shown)


rmarrowstone (approved):

amazon/ion/ioncmodule.c
```diff
@@ -420,14 +420,23 @@ static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_
...
-    IONCHECK(ion_int_from_string(&ion_int_value, &string_value));
-    IONCHECK(ion_writer_write_ion_int(writer, &ion_int_value));
+    PyObject* int_str = NULL;
+    int overflow;
+    long long int_value = PyLong_AsLongLongAndOverflow(obj, &overflow);
+
+    if (!overflow && PyErr_Occurred() == NULL) {
+        // Value fits within int64, write it as int64
+        IONCHECK(ion_writer_write_int64(writer, int_value));
+    } else{
```
tgregg, line 430:
```suggestion
    } else {
```

tgregg (approved):
Nice!

PR #322, opened 2023-12-20
Added and Plumbed IonPyValueModel Flags

rmarrowstone:
This change adds an IonPyValueModel IntEnum to allow users to control
how Ion values map to IonPy and/or standard Python types when reading.

As noted previously, emitting "bare" values is significantly faster
than emitting IonPy values. I also added flags to affect how Symbols
and Structs map, though they are not yet supported. Profiling revealed
that using Python std types for these two had the most impact on perf,
aside from Timestamps which we are tackling independently.

My plan is to make these two flags work in the c-extension, evaluate
and go from there.

**Reviewer's Note:**
I put this into CR as-is to get some feedback on the naming and semantics
of the flags. I don't intend to merge any flags exposed to the user that aren't 
implemented, at least in the c-extension.

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -17,6 +17,8 @@
 The below table describes how types from the Ion data model map to the IonPy types
 in simple_types.py as well as what other Python types are supported on dump.
 
+TODO: add "bare" mapping to table.
+
         +-------------------+-------------------+-----------------------------------+
         | Ion Data Type     | IonPy Type        | Other Dump Mappings               |
         |-------------------+-------------------|-----------------------------------|
@@ -56,7 +58,7 @@ import io
 import warnings
 from datetime import datetime
 from decimal import Decimal
-from enum import IntEnum
+from enum import IntFlag
 from io import BytesIO, TextIOBase
 from types import GeneratorType
 from typing import Union
@@ -156,7 +158,7 @@ def dumps(obj, imports=None, binary=True, sequence_as_stream=False,
     return ret_val
 
 
-class IonPyValueModel(IntEnum):
+class IonPyValueModel(IntFlag):
     """Flags to control the types of values that are emitted from load(s).
 
     The flags may be mixed so that users can intentionally demote certain Ion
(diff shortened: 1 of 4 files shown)


amazon/ion/simpleion.py
```diff
@@ -155,7 +156,56 @@ def dumps(obj, imports=None, binary=True, sequence_as_stream=False,
...
+
+    MIXED = 1
+    """Values will be of the IonPy classes when needed to ensure data fidelity,
+    otherwise they will be standard Python types.
+    """
+
+    SYMBOL_AS_TEXT = 2
+    """Symbol values will be IonPyText(IonType.SYMBOL) or str if bare.
+    
+    Symbol Ids are always lost if present. When bare values are emitted, the
```
tgregg, line 192: An error will have to be raised if the symbol has unknown text (the only case where the symbol ID (really, the import location)) matters.

amazon/ion/simpleion.py
```diff
@@ -155,7 +156,56 @@ def dumps(obj, imports=None, binary=True, sequence_as_stream=False,
...
+        SEXP_AS______ (list, tuple, either?)
+        IGNORE_ANNOTATIONS
+        TIMESTAMP_AS_DATETIME
+        ALWAYS_STD (union of all flags)
+    """
+
+    ION_PY = 0
+    """All values will be instances of the IonPy classes."""
+
+    MIXED = 1
```
tgregg, line 184: Maybe this one could be more descriptive? e.g. `BARE_WHEN_POSSIBLE` or something...

tgregg (approved):
I think this does illustrate that a simple boolean isn't sufficient. I could imagine wanting my "symbols as text", while having a use case that required supporting duplicate field names in structs.

amazon/ion/simpleion.py
rmarrowstone: Yes! Or something like it. I actually have a revision in my workspace that says this:
```
    MIXED = 1
    """Values will be of the IonPy classes when needed to ensure data fidelity,
    otherwise they will be standard Python types.
    
    todo: I'm not sure about the name or the description... this represents
    the legacy "bare" values boolean. But what _is_ a "bare" value? for some
    types that is the standard Python type. But for a Symbol or a Timestamp
    that is a type defined in amazon.ion.core that is _not_ a "IonPy" type.
    And for struct there is no "bare" type, at least currently.
    """
```

amazon/ion/simpleion.py
rmarrowstone: FTR we're already losing strict fidelity for this case, which is I suppose my rebuttal to your comment on https://github.com/amazon-ion/ion-python/issues/316

rmarrowstone: I see. That currently raises an error as well, though we could be returning a SymbolToken with None text and Sid and import location.

PR #323, opened 2023-12-21
Cache the attribute name that is accessed repeatedly.

linlin-s:
*Issue #, if available:*

N/A

*Description of changes:*

This PR introduces caching for the attribute names` ion_type` and `ion_annotations`. The motivation for this change comes from the [implementation ](https://github.com/python/cpython/blob/1ff02385944924db7e683a607da2882462594764/Objects/object.c#L1016) of `PyObject_GetAttrString()`, which includes a step to convert a C-style string to a Python string. To bypass this conversion process for frequently accessed attribute names like ` ion_type` and `ion_annotations`, we now cache these attribute names. In this case, we can use `PyObject_GetAttr()` directly for attribute access, enhancing overall efficiency.

Here are the benchmark results from benchmarking ion-python writing using service_log_legacy: 
There is _**21.64%**_ performance improvement from the change.

To reproduce this results, please run: 
`python ion_benchmark_cli.py write --iterations 30 --warmups 10 --io-type fileservice_log_legacy.ion --format ion_binary`

| Before |       21270622 |  2628522991.60 |   2637737095.84 |               42664252 |
| After |       21270622 |  2056955866.80 |   2066759645.84 |               42593533 |

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -132,7 +132,7 @@ static PyTypeObject ionc_read_IteratorType = {
  *  Returns:
  *      An attribute as an int
  */
-static int int_attr_by_name(PyObject* obj, int* attr_name) {
+static int int_attr_by_name(PyObject* obj, PyObject* attr_name) {
     PyObject* py_int = PyObject_GetAttr(obj, attr_name);
     int c_int = 0;
     if (py_int != Py_None) {
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -72,6 +72,8 @@ static PyObject* _exception_module;
 static PyObject* _ion_exception_cls;
 static PyObject* _add_item;
 static decContext dec_context;
+static PyObject* ion_type_str;
+static PyObject* ion_annotations_str;
 
 typedef struct {
     PyObject *py_file; // a TextIOWrapper-like object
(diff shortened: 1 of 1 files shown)


tgregg (approved):
Wow, that's a big improvement for such a simple change. Do you plan to apply this everywhere we are using PyObject_GetAttrString with known strings? I see usages in `ionc_write_symboltoken` and several in `ionc_write_value` for timestamps (including via `int_attr_by_name`).

linlin-s:
Here is the new benchmark results after the commit: There is _**26.72%**_ performance improvement from the change.
| name                                      |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
| Before |       21270622 |  2609352666.60 |   2618688976.11 |               42637889 |
| After |       21270622 |  1903635733.40 |   1918800344.15 |               42622688 |

linlin-s:
> Wow, that's a big improvement for such a simple change. Do you plan to apply this everywhere we are using PyObject_GetAttrString with known strings? I see usages in `ionc_write_symboltoken` and several in `ionc_write_value` for timestamps (including via `int_attr_by_name`).

Thanks for reminding. I was plan to apply this change on` ionc_write_symboltoken` but overlook the method `int_attr_by_name`. After apply this change to both of them, there is total  **_26.72%_** performance improvements on `service_log_leagcy` data.

amazon/ion/ioncmodule.c
```diff
@@ -116,8 +132,8 @@ static PyTypeObject ionc_read_IteratorType = {
  *  Returns:
  *      An attribute as an int
  */
-static int int_attr_by_name(PyObject* obj, char* attr_name) {
-    PyObject* py_int = PyObject_GetAttrString(obj, attr_name);
+static int int_attr_by_name(PyObject* obj, int* attr_name) {
```
tgregg, line 135: Why is this `int*` and not `PyObject*`?

linlin-s: > Why is this `int*` and not `PyObject*`?

Thanks for catching this. It should be `PyObject*`, I was following the IDE recommended data type when I was editing this method.

tgregg (approved):
Squash and merge

PR #327, opened 2023-12-28
Implement Symbol as Text in C-extension

rmarrowstone:
By default the IonPy value for a Symbol is IonPySymbol and the bare
value is a SymbolToken. With the SYMBOL_AS_TEXT flag set, the IonPy
value for a Symbol is IonPyText, and the bare value is str.

This makes Symbol handling simpler if one only cares about the text
value. It also improves load performance for symbols significantly.

Symbols with undefined text cannot be emitted with this flag set and
will raise exceptions.

### DIFF
--- a/tests/test_simpleion.py
+++ b/tests/test_simpleion.py
@@ -767,6 +767,10 @@ def test_value_model_flags(params):
 
 
 def test_undefined_symbol_text_as_text():
+    # This function only tests c extension
+    if not c_ext:
+        return
+
     ion_text = """
     $ion_symbol_table::{ symbols:[ null ] }
     $10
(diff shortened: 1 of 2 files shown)


tests/test_simpleion.py
```diff
@@ -741,22 +741,47 @@ def test_bare_values(params):
...
+        ion_text, value_model, expected_type, expected_ion_type = params
 
-    try:
+    value = simpleion.load_extension(StringIO(ion_text), value_model=value_model)
+    assert type(value) is expected_type
+    if expected_ion_type:
+        assert value.ion_type == expected_ion_type
+
+
+def test_undefined_symbol_text_as_text():
```
tgregg, line 769: Looks like this test might need
```
    # This function only tests c extension
    if not c_ext:
        return
```
So that it doesn't fail on Pypy.

tgregg (approved):

tests/test_simpleion.py
rmarrowstone: Oof yea.

PR #334, opened 2024-01-08
Optimize ionc_write_struct

linlin-s:
*Issue #, if available:*

*Description of changes:*

This PR optimized the handling of `IonPyDict` objects. In the previous implementation, we converted the `PyObject` map into a `sequence` for iteration. The updated approach introduced in this PR separates the handling into two cases based on the type of the map object:
1. For Python dictionaries: When `map` is a standard Python dictionary, we iterate over it directly using `PyDict_Next`. 
2. For `IonPyDict` objects: In cases where map is an instance of IonPyDict, we now access its `_ _store` attribute. This attribute itself is a Python dictionary. By doing so, we can then use `PyDict_Next` to iterate over __store, similar to how we handle standard Python dictionaries.

This optimization eliminates the need to convert IonPyDict objects into sequences before iteration, leading to a more efficient processing process. 

Here are the benchmark results from benchmarking ion-python writing using service_log_legacy:
 There is **_31.2%_** performance improvement from the change.

To reproduce this results, please run:
`python ion_benchmark_cli.py write --iterations 30 --warmups 10 --io-type fileservice_log_legacy.ion --format ion_binary`

| name                                      |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
| Before |       21270622 |  1715543333.40 |   1753941142.77 |               42597119 |
| After|       21270622 |  1174766083.40 |   1206556276.95 |               42582427 |

### DIFF
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -91,6 +91,7 @@ static PyObject* hour_str;
 static PyObject* minute_str;
 static PyObject* second_str;
 static PyObject* microsecond_str;
+static PyObject* store_str;
 
 typedef struct {
     PyObject *py_file; // a TextIOWrapper-like object
@@ -416,16 +417,14 @@ static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_
             IONCHECK(write_struct_field(writer, key, val, tuple_as_sexp));
         }
     } else {
-        store = PyObject_CallMethod(map, "_get_store", NULL);
+        store = PyObject_GetAttr(map, store_str);
         if (store == NULL || !PyDict_Check(store)) {
-            IONCHECK(IERR_INVALID_ARG);
-            goto fail;
+            _FAILWITHMSG(IERR_INVALID_ARG, "Failed to retrieve 'store': Object is either NULL or not a Python dictionary.");
         }
         pos = 0;
         while (PyDict_Next(store, &pos, &key, &val_list)) {
             if (!PyList_Check(val_list)) {
-                IONCHECK(IERR_INVALID_ARG);
-                goto fail;
+                _FAILWITHMSG(IERR_INVALID_ARG, "Invalid value type for the key: Expected a list, but found a different type.");
             }
             list_len = PyList_Size(val_list);
             for (i = 0; i < list_len; i++) {
@@ -436,10 +435,7 @@ static iERR ionc_write_struct(hWRITER writer, PyObject* map, PyObject* tuple_as_
         Py_DECREF(store);
     }
 
-    fail:
-    if (PyErr_Occurred()) {
-        IONCHECK(IERR_INVALID_ARG);
-    }
+fail:
     cRETURN;
 }
 
@@ -1704,6 +1700,7 @@ PyObject* ionc_init_module(void) {
     minute_str = PyUnicode_FromString("minute");
     second_str = PyUnicode_FromString("second");
     microsecond_str = PyUnicode_FromString("microsecond");
+    store_str = PyUnicode_FromString("_IonPyDict__store");
 
     return m;
 }
--- a/amazon/ion/ioncmodule.c
+++ b/amazon/ion/ioncmodule.c
@@ -372,6 +372,34 @@ fail:
     Py_DECREF(sequence);
     cRETURN;
 }
+/*
+ * Process and write the key-value pair.
+ *  *  Args:
+ *      writer:  An ion writer
+ *      key: The key of IonStruct item
+ *      val: The value of IonStruct item
+ *      tuple_as_sexp: Decides if a tuple is treated as sexp
+ */
+
+static iERR write_key_value(hWRITER writer, PyObject* key, PyObject* val, PyObject* tuple_as_sexp) {
+    iERR err;
+    if (PyUnicode_Check(key)) {
+        ION_STRING field_name;
+        ion_string_from_py(key, &field_name);
+        IONCHECK(ion_writer_write_field_name(writer, &field_name));
+    } else if (key == Py_None) {
+        IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
+    }
+    IONCHECK(Py_EnterRecursiveCall(" while writing an Ion struct"));
+    err = ionc_write_value(writer, val, tuple_as_sexp);
+    Py_LeaveRecursiveCall();
+    IONCHECK(err);
+
+    return IERR_OK;
+
+    fail:
+    return err;
+}
 
 /*
  *  Writes a struct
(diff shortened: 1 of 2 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -372,6 +372,34 @@ static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tu
...
+ */
+
+static iERR write_key_value(hWRITER writer, PyObject* key, PyObject* val, PyObject* tuple_as_sexp) {
+    iERR err;
+    if (PyUnicode_Check(key)) {
+        ION_STRING field_name;
+        ion_string_from_py(key, &field_name);
+        IONCHECK(ion_writer_write_field_name(writer, &field_name));
+    } else if (key == Py_None) {
+        IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
```
rmarrowstone, line 391: Is there not a public API for this? I know we are the developers of ion-c but I don't think we want to be calling `_` prefixed methods from another repo?

amazon/ion/ioncmodule.c
```diff
@@ -372,6 +372,34 @@ static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tu
...
+/*
+ * Process and write the key-value pair.
+ *  *  Args:
+ *      writer:  An ion writer
+ *      key: The key of IonStruct item
+ *      val: The value of IonStruct item
+ *      tuple_as_sexp: Decides if a tuple is treated as sexp
+ */
+
+static iERR write_key_value(hWRITER writer, PyObject* key, PyObject* val, PyObject* tuple_as_sexp) {
```
rmarrowstone, line 384: nit: name could be clearer. I would call it `write_key_value_pair` or just `write_struct_field` but I read it as "write the value of the key" maybe just me though.

amazon/ion/ioncmodule.c
```diff
@@ -372,6 +372,34 @@ static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tu
...
+        IONCHECK(ion_writer_write_field_name(writer, &field_name));
+    } else if (key == Py_None) {
+        IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
+    }
+    IONCHECK(Py_EnterRecursiveCall(" while writing an Ion struct"));
+    err = ionc_write_value(writer, val, tuple_as_sexp);
+    Py_LeaveRecursiveCall();
+    IONCHECK(err);
+
+    return IERR_OK;
```
rmarrowstone, line 398: If you're using `IONCHECK` I think there is an `iReturn` or some other macro that would be better. Don't have code open so IDK.

amazon/ion/simple_types.py
```diff
@@ -744,6 +744,12 @@ def items(self):
         """
         return [i for i in self.iteritems()]
 
+    def get_store(self):
```
rmarrowstone, line 747: Add an `_` to suggest privateness please.

amazon/ion/ioncmodule.c
```diff
@@ -384,48 +412,37 @@ static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tu
...
-        Py_INCREF(key);
-        Py_INCREF(val);
-
-        if (PyUnicode_Check(key)) {
-            ION_STRING field_name;
-            ion_string_from_py(key, &field_name);
-            IONCHECK(ion_writer_write_field_name(writer, &field_name));
+    PyObject *store = NULL, *key = NULL, *val_list = NULL, *val = NULL;
+    Py_ssize_t pos = 0, i, list_len;
+    if (PyDict_Check(map)) {
```
rmarrowstone, line 417: Please ensure we have test coverage for `IonPyDict`, `IonPyStdDict` and std `dict`s.

amazon/ion/ioncmodule.c
linlin-s: > nit: name could be clearer. I would call it `write_key_value_pair` or just `write_struct_field` but I read it as "write the value of the key" maybe just me though.

Thanks for calling this out, `write_struct_field` works better in this case.

amazon/ion/ioncmodule.c
linlin-s: This concern deserves more investigation. Unfortunately I couldn't find the public API for this functionality in `ion-c`. There are several spots using the `_` prefixed methods in [ionc_write_symboltoken](https://github.com/amazon-ion/ion-python/blob/d425587d94f8e9fa22368ed1b82a1b65fcb81820/amazon/ion/ioncmodule.c#L278C18-L278C18). I'm considering whether we should propose a new strategy for these implementations that avoids using the `_` prefixed methods, and then update them all at once.

amazon/ion/ioncmodule.c
linlin-s: > If you're using `IONCHECK` I think there is an `iReturn` or some other macro that would be better. Don't have code open so IDK.

Yeah that's a good catch. Here is how we define iRETURN:
` #define iRETURN       fail: RETURN(__location_name__, __line__, __count__++, err)` .
 In this case we can replace `return IERR_OK; `with `iRETURN`.

amazon/ion/ioncmodule.c
rmarrowstone: Well ok, but that's not great. I also see this pre-existed...

But really this is papering over a potential data correctness issue when roundtripping Ion with unresolved imports. Should probably just throw here. I know that's technically a "breaking change" but we should've just thrown here in the first place (aside from implementing the round-tripping correctly).

amazon/ion/ioncmodule.c
rmarrowstone: Users should be able to write a standard python `dict` object as an Ion struct, and likewise an `IonPyStdDict` (which they _should_ be able to given that it inherits from `dict`). I do not know if there are existing tests or not.

amazon/ion/ioncmodule.c
rmarrowstone: ... on further consideration:
There is a correctness issue with roundtripping, but that needs to be handled on the read side.

The way to do this without calling protected/private methods is to construct an `ION_SYMBOL` that has null text and SId of 0 and no import location. Of course this is pre-existing so you're not obligated to.

linlin-s: Thanks for the explanation.

amazon/ion/ioncmodule.c
linlin-s: Thanks for the clarification. We do have test for [IonPyDict](https://github.com/amazon-ion/ion-python/blob/d425587d94f8e9fa22368ed1b82a1b65fcb81820/tests/test_simple_types.py#L59) and [dict](https://github.com/amazon-ion/ion-python/blob/d425587d94f8e9fa22368ed1b82a1b65fcb81820/tests/test_simple_types.py#L115). For `IonPyStdDict`, it should be covered in this [test](https://github.com/amazon-ion/ion-python/blob/d425587d94f8e9fa22368ed1b82a1b65fcb81820/tests/test_simple_types.py#L115) since it's subclass of `dict`.

amazon/ion/ioncmodule.c
```diff
@@ -384,48 +409,37 @@ static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tu
...
+            IONCHECK(write_struct_field(writer, key, val, tuple_as_sexp));
         }
-        else if (key == Py_None) {
-            // if field_name is None, write symbol $0 instead.
-            IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
+    } else {
+        store = PyObject_CallMethod(map, "_get_store", NULL);
+        if (store == NULL || !PyDict_Check(store)) {
+            IONCHECK(IERR_INVALID_ARG);
+            goto fail;
```
rmarrowstone, line 422: Prefer `FAILWITH(IERR_INVALID_ARG)` or consider `_FAILWITHMSG(...)`.

Yes! I know I encouraged you to avoid `_` methods (in this case a macro) but for reasons I don't understand `_FAILWITHMSG` actually propagates the message back to the Python exception whereas the non-underscored doesn't.

amazon/ion/ioncmodule.c
```diff
@@ -384,48 +409,37 @@ static iERR ionc_write_sequence(hWRITER writer, PyObject* sequence, PyObject* tu
...
+    Py_ssize_t pos = 0, i, list_len;
+    if (PyDict_Check(map)) {
+        while (PyDict_Next(map, &pos, &key, &val)) {
+            IONCHECK(write_struct_field(writer, key, val, tuple_as_sexp));
         }
-        else if (key == Py_None) {
-            // if field_name is None, write symbol $0 instead.
-            IONCHECK(_ion_writer_write_field_sid_helper(writer, 0));
+    } else {
+        store = PyObject_CallMethod(map, "_get_store", NULL);
```
rmarrowstone, line 419: Just an FYI that you could do: `PyObject_GetAttr(map, "_IonPyDict__store")`

rmarrowstone (approved):

amazon/ion/ioncmodule.c
nirosys: > Just an FYI that you could do: `PyObject_GetAttr(map, "_IonPyDict__store")`

This caught my eye, so I found the relevant documentation for anyone else who's curious: [Private name mangling](https://docs.python.org/3/reference/expressions.html#atom-identifiers)

amazon/ion/ioncmodule.c
nirosys: Dug into this a little..

`FAILWITHMSG` is provided by ion-c, but`_FAILWITHMSG` is defined in [ioncmodule.c](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/ioncmodule.c#L28), and copies the provided error msg into the global `_err_msg` . `ionc_read` and `ionc_write` (and other functions like iter read) use `_err_msg` when setting the error indicator (like [here](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/ioncmodule.c#L861)).

linlin-s: Thanks for the suggestion. In this case I would choose to use  `_FAILWITHMSG(...)` to align with the err handling in other writing implementations.

linlin-s: > Dug into this a little..
> 
> `FAILWITHMSG` is provided by ion-c, but`_FAILWITHMSG` is defined in [ioncmodule.c](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/ioncmodule.c#L28), and copies the provided error msg into the global `_err_msg` . `ionc_read` and `ionc_write` (and other functions like iter read) use `_err_msg` when setting the error indicator (like [here](https://github.com/amazon-ion/ion-python/blob/master/amazon/ion/ioncmodule.c#L861)).

Thanks for the information!

amazon/ion/ioncmodule.c
linlin-s: This is a better strategy to get `_ _store` attribute. By avoiding calling `PyObject_CallMethod(map, "_get_store", NULL);`, we gained another _**9%**_ performance improvements! After rerun the benchmark before and after changes, there is overall **_40%_** improvement on writing `service_log_legacy` data.
| name                                      |   file_size(B) |   time_min(ns) |   time_mean(ns) |   memory_usage_peak(B) |
| Before |       21270622 |  1737683900.00 |   1848320107.78 |               42584810 |
| After |       21270622 |  1059837258.20 |   1107548862.48 |               42571174 |

rmarrowstone (approved):

PR #336, opened 2024-01-12
Use PyDateTime C API to Write Timestamps

rmarrowstone:
This change uses the PyDateTime C API to access the Timestamp
attributes from the datetime super class. It improves Timestamp
serialization by about 10% for both Text and Binary. I expect the
improvement for basic datetimes to be higher but I haven't measured
it.

It should be compatible with future changes to the Timestamp types
so long as they still inherit from DateTime, which I think is a safe
assumption.

*Issue #, if available:*

*Description of changes:*


tgregg (approved):
Can the timestamp-specific strings (e.g. `month_str`, `day_str`) be cleaned up now too, or are they used elsewhere?

rmarrowstone:
> Can the timestamp-specific strings (e.g. `month_str`, `day_str`) be cleaned up now too, or are they used elsewhere?

Addressed.

PR #337, opened 2024-02-13
Optimize Timestamp Reads in ion-c extension

rmarrowstone:
The first commit refactors Timestamp construction, pulling the heavy-lifting validation and attribute setting logic out into
__init__ and leaving the bare minimum in __new__.

The second commit calls __new__ directly from ionc_read_timestamp and while avoiding building a dict of keyword
args.

A benchmark test I put together of all the timestamps from various "vectors" tests shows that this is about 2.7x faster
for reading timestamps. The impact to my "hkc" ion benchmark, which features a single Timestamp out of about 15 attributes per top-level-value is roughly a 25% improvement.

### DIFF
--- a/amazon/ion/core.py
+++ b/amazon/ion/core.py
@@ -11,8 +11,10 @@
 # OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the
 # License.
+import warnings
 from enum import IntEnum
 from typing import NamedTuple, Optional, Any, Union, Sequence, Coroutine
+from warnings import warn
 
 from amazon.ion.symbols import SymbolToken
 
@@ -326,7 +328,10 @@ _ZERO_DELTA = timedelta()
 
 
 class OffsetTZInfo(tzinfo):
-    """A trivial UTC offset :class:`tzinfo`."""
+    """A trivial UTC offset :class:`tzinfo`.
+
+    todo: as of Python 3 this can be dropped in favor of datetime.timezone(offset_delta)
+    """
     def __init__(self, delta=_ZERO_DELTA):
         if delta <= _MIN_OFFSET or delta >= _MAX_OFFSET:
             raise ValueError('Invalid UTC offset: %s' % delta)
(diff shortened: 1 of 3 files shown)


amazon/ion/ioncmodule.c
```diff
@@ -900,122 +902,89 @@ static PyObject* ionc_get_timestamp_precision(int precision) {
...
 
-                PyObject* py_microsecond = PyLong_FromLong(microsecond);
-                PyObject* py_fractional_precision = PyLong_FromLong(fractional_precision);
-                PyDict_SetItemString(timestamp_args, "microsecond", py_microsecond);
-                PyDict_SetItemString(timestamp_args, "fractional_precision", py_fractional_precision);
-                Py_DECREF(py_microsecond);
-                Py_DECREF(py_fractional_precision);
-            }
+            if (fractional_precision > MICROSECOND_DIGITS) fractional_precision = MICROSECOND_DIGITS;
+            py_fractional_seconds = PyObject_CallFunction(_decimal_constructor, "s", dec_num, NULL);
```
tgregg, line 953: Are you asserting that using a string as the intermediary between decQuad and Decimal is faster than construction using coefficient/scale, or is it just less code? Either way, it's interesting.

tgregg (approved):

amazon/ion/ioncmodule.c
rmarrowstone: I was just looking at how to simplify. To make sure I understand your question though: we were already doing decQuad => string => Decimal for the arbitrary precision path, I just collapsed the "micros or less precise" path to that.

I did that because the `Timestamp.__new__` _must_ derive the microseconds if they are not provided (due to the python requirement that the `datetime` fields be set in `__new__` and not `__init__`). So for the "micros or less precise" path I would need to do the `fractional_seconds` and `py_fractional_precision` here anyway (deriving them in python is slow so I kept that out of `__new__`).

So, given that I had to get the `fractional_seconds` and `py_fractional_seconds` for either path here, _and_ had to derive `microsecond` in `__new__`, this seemed like the right mix of simplicity to performance.

tgregg: Makes sense, thanks.

amazon/ion/core.py
```diff
@@ -11,8 +11,10 @@
 # OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the
 # License.
+import warnings
```
rmarrowstone, line 14: Just noticed this, will remove.

amazon/ion/core.py
```diff
@@ -11,8 +11,10 @@
...
 from enum import IntEnum
 from typing import NamedTuple, Optional, Any, Union, Sequence, Coroutine
+from warnings import warn
```
rmarrowstone, line 17: ditto

nirosys (approved):

PR #338, opened 2024-02-13
Updates the release workflow to support python 3.11 and MacOS arm64.

cheqianh:
## Description 

This PR adds support for wheels on Python 3.11 and MacOS-arm64.

In detail, this PR:
1. Adds wheels for Python 3.11.
2. Adds wheels for MacOS-arm64.
3. Drops support for wheels built with Python 3.6 to 3.8.
4. Updates required dependencies and build tools.

## Test

1. A successful wheel build workflow from my fork can be found at: https://github.com/cheqianh/ion-python/actions/runs/7880967937.
2. All tests have passed on GitHub Actions 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,64 +25,64 @@ jobs:
       - run: pip install -e .
       - run: py.test
 
-#  source-distribution:
-#    if: startsWith(github.ref, 'refs/tags/')
-#    name: Source distribution
-#    runs-on: ubuntu-latest
-#    needs: [test]
-#    strategy:
-#      matrix:
-#        python-version: [ '3.10' ]
-#    steps:
-#      - uses: actions/checkout@v3
-#
-#      - name: Build sdist
-#        run: python3 setup.py sdist
-#
-#      - name: Upload source to Github
-#        uses: actions/upload-artifact@v2
-#        with:
-#          name: source
-#          path: ./dist/amazon.ion-*.tar.gz
-#
-#      - uses: aws-actions/configure-aws-credentials@v1
-#        with:
-#          role-skip-session-tagging: true
-#          aws-region: us-west-2
-#          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-#          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-#          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
-#          role-external-id: ${{ secrets.AWS_ROLE_EXTERNAL_ID }}
-#          role-duration-seconds: 900
-#
-#      - name: Upload source to s3
-#        run: |
-#          zip ion-python-source ./dist/amazon.ion-*.tar.gz
-#          aws s3 mv ion-python-source.zip ${{ secrets.AWS_SOURCE_BUCKET_URL }} --acl bucket-owner-full-control
-#
-#      - name: Install the released source from PYPI
-#        run: |
-#          sleep 300s
-#
-#          for (( i=0; i<3; i++ ))
-#          do
-#            v=""
-#
-#            if ! (pip install --no-binary :all: amazon.ion==${GITHUB_REF##*/v}) then
-#              echo "Unable to install the desired version"
-#              sleep 120s
-#              continue
-#            fi
-#
-#            v=$( pip show amazon.ion | grep Version )
-#            break
-#          done
-#
-#          if [[ $v != "Version: ${GITHUB_REF##*/v}" ]]
-#          then
-#            echo "Exiting because unable to install the new version from PYPI"
-#            exit 1
-#          fi
+  source-distribution:
+    if: startsWith(github.ref, 'refs/tags/')
+    name: Source distribution
+    runs-on: ubuntu-latest
+    needs: [test]
+    strategy:
+      matrix:
+        python-version: [ '3.10' ]
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -91,11 +92,26 @@ jobs:
...
+            cibw_arch: "auto aarch64"
+
     steps:
       - uses: actions/checkout@v3
 
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+
```
rmarrowstone, line 114: What is the relationship between this python 3.10 setup step and the built wheels? Is 3.10 used to build wheels for all the versions? Are we only building wheels for 3.10?

rmarrowstone (approved):
I realize that we were already using 3.10 for the python version for building the wheels, but let's make sure we understand what's going on there.

.github/workflows/release.yml
tgregg: And why not 3.11?

cheqianh: This python version will be used for setting up virtual environment and install dependencies including `cibuildwheel`. We used `3.10` back then because `cibuildwheel=2.4.0` only supported wheels for python versions up `3.10` so I used the same version as it's the latest version of python we supported to build wheels. 

To avoid the confusion, I updated the PR to use general `python3` command to let tasks use the default python versions that these images are using. For example, [MacOs 14](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md) is using python `3.11` and [Ubuntu-latest](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) is using python `3.10`. But both versions are tested in our [main test workflow](https://github.com/amazon-ion/ion-python/pull/338/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R33).

cheqianh: A successful workflow without python 3.10 setup is available in my fork - https://github.com/cheqianh/ion-python/actions/runs/7891635687

.github/workflows/release.yml
```diff
@@ -108,15 +119,13 @@ jobs:
...
+        run: python3 -m pip install cibuildwheel==2.16.5
 
       - name: Build wheels
         run: python3 -m cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_SKIP: "pp* *-win32"
-          CIBW_ARCHS_WINDOWS: "auto"
-          CIBW_ARCHS_LINUX: "auto aarch64"
-          CIBW_ARCHS_MACOS: "x86_64"
+          CIBW_SKIP: "pp* *-win32 cp36-* cp37-* cp38-*"
```
rmarrowstone, line 127: Prefer explicit including instead of accepting the `cibuildwheel` defaults and skipping some. That way we _know_ that we are only building for arch/python version combinations that we have tested for and it won't change with `cibuildwheel`.

.github/workflows/release.yml
rmarrowstone: Well that still doesn't really answer the question. I dug into the docs and the answer is that the `CIBW` flags control  which architectures and python versions one builds for. I added a comment to the current skip setup because I think we need to be explicit about we _include_ not _exclude_ there. It would also make it clearer what is going on.

.github/workflows/release.yml
rmarrowstone: Please also add a comment about what these apparently arcane flags mean for the next dev that has to change this.

.github/workflows/release.yml
cheqianh: Okay, will include this change in the next commit.

cheqianh: >  Is 3.10 used to build wheels for all the versions? 

No, `cibuildwheel` will build wheels using different python versions respectively  

> Are we only building wheels for 3.10?

We are currently supporting the building of wheels for versions 3.8 to 3.11.

.github/workflows/release.yml
cheqianh: Added a commit to
* uses specific arch names instead of alias "auto",
* uses specific python versions 3.9 to 3.11
* Adds a link to `cibuildwheel` document

I tried to add python 3.12 but the build failed due to a dependency issue. I'll open a new issue to track the 3.12 build issue.

cheqianh: Opened a new issue for adding 3.12 to the build workflow - https://github.com/amazon-ion/ion-python/issues/339

.github/workflows/release.yml
```diff
@@ -108,15 +120,13 @@ jobs:
...
+        run: python3 -m pip install cibuildwheel==2.16.5
 
       - name: Build wheels
         run: python3 -m cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_SKIP: "pp* *-win32"
-          CIBW_ARCHS_WINDOWS: "auto"
-          CIBW_ARCHS_LINUX: "auto aarch64"
-          CIBW_ARCHS_MACOS: "x86_64"
+          CIBW_BUILD: "cp39-* cp310-* cp311-*"
```
rmarrowstone, line 128: Please add comment. It is really non-obvious that "cp39" means "C-Python 3.9".

rmarrowstone (approved):

.github/workflows/release.yml
rmarrowstone: I mean... until it is, of course! ;-)

PR #341, opened 2024-03-29
Refactor simpleion pure-python load methods

rmarrowstone:
This change removes some duplication in the simpleion implementation.

Before this change the _load and _load_iteratively methods were near
duplicates and used the python call stack for maintaining the Ion
container stack.

After this change _load is removed in favor of _load_iteratively which
uses a deque to maintain the container stack.

It also factors out the handling of the single_value and parse_eagerly
flags which was also nearly duplicated between the ionc and
pure-python usages.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -56,6 +56,7 @@ disabled as below:
 """
 import io
 import warnings
+from collections import deque
 from datetime import datetime
 from decimal import Decimal
 from enum import IntFlag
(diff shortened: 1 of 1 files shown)


tgregg (commented):
This approach looks good to me. Any idea if there is performance impact?

rmarrowstone:
> This approach looks good to me. Any idea if there is performance impact?
There is a significant regression for pypy. I'm not actually concerned about _that_, but I want to gather more data about cpython before moving forward, so i'm postponing this for now.

rmarrowstone:
See comment about pypy. Will re-open if/when I see enough benefit (or at least not harm) on cpython.

PR #342, opened 2024-04-03
Add dependabot configuration for keeping dependencies current

zachborboacryptofi:
This change adds a Dependabot[1] configuration to keep dependencies up-to-date. The routine will check files like https://github.com/amazon-ion/ion-python/blob/master/requirements.txt for any updates.

[1] https://docs.github.com/en/code-security/dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


popematt (approved):
Thanks for your contribution. This is a good idea, and something that we've overlooked for far too long due to other priorities.

I've changed the configured schedule so that it's not as noisy (only create a PR once a week, and group all dependency updates together), and I've also added in configuration for updating the GitHub Action dependencies too.

zachborboacryptofi:
one option is to temporarily ungroup the updates until they're mostly updated. could help move https://github.com/amazon-ion/ion-python/pull/344 forward

PR #372, opened 2024-10-10
feat: Add `trailing_commas` to `dump`/`dumps`

mavjop:
*Issue #, if available:*
[Issue # 370](https://github.com/amazon-ion/ion-python/issues/370)

*Description of changes:*

Add an optional boolean argument `trailing_commas` to the `dump` and `dumps` methods in `simpleion`. Default: False.

When `trailing_commas=True`, and when `indent` is not `None`, this causes `dump[s]` to include a comma on the last item in a container.

**Note:** Since the C module already **does not support pretty printing**, this flag does nothing when using the C module.

### DIFF
--- a/amazon/ion/simpleion.py
+++ b/amazon/ion/simpleion.py
@@ -95,7 +95,7 @@ c_ext = __IS_C_EXTENSION_SUPPORTED
 
 
 def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, indent=None,
-         tuple_as_sexp=False, omit_version_marker=False):
+         tuple_as_sexp=False, omit_version_marker=False, trailing_commas=False):
     """Serialize ``obj`` as an Ion formatted stream and write it to fp.
 
     The python object hierarchy is mapped to the Ion data model as described in the module pydoc.
@@ -121,11 +121,15 @@ def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, indent=No
         indent (Str): If binary is False and indent is a string, then members of containers will be pretty-printed with
             a newline followed by that string repeated for each level of nesting. None (the default) selects the most
             compact representation without any newlines. Example: to indent with four spaces per level of nesting,
-            use ``'    '``.
+            use ``'    '``. Supported only in the pure Python implementation (because pretty printing is not yet
+            supported in the C implementation).
         tuple_as_sexp (Optional[True|False]): When True, all tuple values will be written as Ion s-expressions.
             When False, all tuple values will be written as Ion lists. Default: False.
         omit_version_marker (Optional[True|False]): If binary is False and omit_version_marker is True, omits the
             Ion Version Marker ($ion_1_0) from the output.  Default: False.
+        trailing_commas (Optional[True|False]): If binary is False and pretty printing (indent is not None), includes
+            trailing commas in containers. Default: False. Supported only in the pure Python implementation (because
+            pretty printing is not yet supported in the C implementation).
 
     Returns None.
     """
@@ -135,11 +139,12 @@ def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, indent=No
     else:
         return dump_python(obj, fp, imports=imports, binary=binary, sequence_as_stream=sequence_as_stream,
                            indent=indent,
-                           tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker)
+                           tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker,
+                           trailing_commas=trailing_commas)
 
 
 def dumps(obj, imports=None, binary=True, sequence_as_stream=False,
-          indent=None, tuple_as_sexp=False, omit_version_marker=False):
+          indent=None, tuple_as_sexp=False, omit_version_marker=False, trailing_commas=False):
     """Serialize obj as described by dump, return the serialized data as bytes or unicode.
 
     Returns:
@@ -149,7 +154,8 @@ def dumps(obj, imports=None, binary=True, sequence_as_stream=False,
     ion_buffer = io.BytesIO()
 
     dump(obj, ion_buffer, imports=imports, sequence_as_stream=sequence_as_stream, binary=binary,
-         indent=indent, tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker)
+         indent=indent, tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker,
+         trailing_commas=trailing_commas)
 
     ret_val = ion_buffer.getvalue()
     ion_buffer.close()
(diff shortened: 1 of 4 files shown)


amazon/ion/simpleion.py
```diff
@@ -121,11 +121,15 @@ def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, indent=No
         indent (Str): If binary is False and indent is a string, then members of containers will be pretty-printed with
             a newline followed by that string repeated for each level of nesting. None (the default) selects the most
             compact representation without any newlines. Example: to indent with four spaces per level of nesting,
-            use ``'    '``.
+            use ``'    '``. Supported only in the pure Python implementation (because pretty printing is not yet
```
rmarrowstone, line 124: Thanks!

amazon/ion/writer_text.py
```diff
@@ -369,7 +369,7 @@ def _raw_writer_coroutine(depth=0, container_event=None, whence=None, indent=Non
         ion_event, self = (yield transition)
         delegate = self
 
-        if has_written_values and not ion_event.event_type.ends_container:
+        if has_written_values and ((indent and trailing_commas) or not ion_event.event_type.ends_container):
```
rmarrowstone, line 372: I may be wrong but I think this will also add a space after the last item in an s-exp: `(foo bar bar )` I'm not sure that's _bad_ but I don't think it's intended.

amazon/ion/simpleion.py
```diff
@@ -121,11 +121,15 @@ def dump(obj, fp, imports=None, binary=True, sequence_as_stream=False, indent=No
...
+            supported in the C implementation).
         tuple_as_sexp (Optional[True|False]): When True, all tuple values will be written as Ion s-expressions.
             When False, all tuple values will be written as Ion lists. Default: False.
         omit_version_marker (Optional[True|False]): If binary is False and omit_version_marker is True, omits the
             Ion Version Marker ($ion_1_0) from the output.  Default: False.
+        trailing_commas (Optional[True|False]): If binary is False and pretty printing (indent is not None), includes
```
rmarrowstone, line 130: Suggestion: it may actually be clearer to a user to make it independent of `indent` I know you don't intend to use it that way, and TBH I don't know the value of doing so, but it would be avoid the need to explain and users understand the dependent relationship, or test for it in the code. 🤷

rmarrowstone (approved):

amazon/ion/writer_text.py
mavjop: Hm, I did think you were right[1], but I actually think not, because there's a separate set of delimiters defined for use when pretty printing:
```
_CONTAINER_DELIMITER_MAP_PRETTY = {
    IonType.STRUCT: b',',
    IonType.LIST: b',',
    IonType.SEXP: b'', # we use newlines when pretty printing
}
```
which uses an empty string as the delimiter for s-expressions, since they add newlines.

If we **do** want to apply this to **non**-pretty-printing, then we **could** do that, but then your concern expressed here **would** need an additional work-around, or there would indeed be an extra space.

I think the solution, if needed, would be to fork:
```
_CONTAINER_DELIMITER_MAP_NORMAL = {
    IonType.STRUCT: b',',
    IonType.LIST: b',',
    IonType.SEXP: b' ',
}
```

into:
```
_CONTAINER_DELIMITER_MAP_NORMAL = {
    IonType.STRUCT: b',',
    IonType.LIST: b',',
    IonType.SEXP: b' ',
}
_CONTAINER_END_DELIMITER_MAP_NORMAL = {
    IonType.STRUCT: b',',
    IonType.LIST: b',',
    IonType.SEXP: b'',
}
```

and use `_CONTAINER_END_DELIMITER_MAP_NORMAL` for the final trailing delimiter.

amazon/ion/simpleion.py
mavjop: That's a thought.

I actually had it accidentally including trailing commas in non-pretty-printed output for the first revision, so it would be easy to put back to that (except for the problem explained in the next paragraph). I guess if you say you don't want indent but you do want trailing commas, there's no strong reason **not** to.

However ... there is one reason not to. This is a reason why I'd rather that be a feature we could consider but not try to include here. That reason is that the Python C module implementation already **doesn't** support pretty-printing, so I was able to leave it out of scope. If we make `trailing_commas` applicable to **non**-pretty-printed output, then I have to go and do all the C module implementation.

I'd rather not hold this PR up to implement non-pretty-printed trailing-comma output in the C module; especially since I'm not sure we're likely to see any strong demand for trailing commas in non-pretty-printed output. I guess we could make it **not** depend on `indent`, but still disclaim it as "not supported in the C implementation". So that's an option.

mavjop:
Ouch, that's a lot of failed checks. :/ I guess I'll have to have a look into those at some point.

amazon/ion/simpleion.py
```diff
@@ -470,10 +477,11 @@ def add(obj):
         event = reader.send(NEXT_EVENT)
 
 
-def dump_extension(obj, fp, binary=True, sequence_as_stream=False, tuple_as_sexp=False, omit_version_marker=False):
+def dump_extension(obj, fp, binary=True, sequence_as_stream=False, tuple_as_sexp=False, omit_version_marker=False,
+                   trailing_commas=False):
     """C-extension implementation. Users should prefer to call ``dump``."""
 
-    res = ionc.ionc_write(obj, binary, sequence_as_stream, tuple_as_sexp)
+    res = ionc.ionc_write(obj, binary, sequence_as_stream, tuple_as_sexp, trailing_commas)
```
nirosys, line 484: This call is what is causing the segfault in the unit tests. `ionc_write` isn't expecting to get the trailing_commas argument, and is not handling it well.

If you want to put in a `# TODO:`, and drop the `trailing_commas` from the `ionc.ionc_write` call, I can dig into why it isn't handling that a bit more gracefully, and that should clear up the test errors (the ones that aren't already explained).

mavjop: Wait, I must be confused. I'm not passing the arg to the C code, as far as I can see?
(Note lack of `trailing_commas=trailing_commas` in the first branch.)

```
    if c_ext and __IS_C_EXTENSION_SUPPORTED and (imports is None and indent is None):
        return dump_extension(obj, fp, binary=binary, sequence_as_stream=sequence_as_stream,
                              tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker)
    else:
        return dump_python(obj, fp, imports=imports, binary=binary, sequence_as_stream=sequence_as_stream,
                           indent=indent,
                           tuple_as_sexp=tuple_as_sexp, omit_version_marker=omit_version_marker,
                           trailing_commas=trailing_commas)
```

mavjop: Oh, oops, in `dump_extension`. D'oh.

mavjop: Done.

(Removed `trailing_commas` from `dump_extension()` and added a `TODO`).

nirosys (approved):
Thank you!

Issue #375, opened 2024-10-18
Invalid operation (segfault, sigbus, etc) when`ionc.ionc_write` is given more arguments than expected.


nirosys:
Identified in #372.

### Issue
```
(venv) glitch@147dda5e5395 ~/C/s/i/ion-python> python
Python 3.12.5 (main, Aug  6 2024, 19:08:49) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from amazon.ion import ionc
>>> ionc.ionc_write(None, True, False, False, None)
fish: Job 1, 'python' terminated by signal SIGBUS (Misaligned address error)
```
Looking at the code, it looks like the issue is the uninitialized PyObject pointers values at the top of `ionc_write`. In the event that the arguments are wrong, the function jumps down to the fail label, and pointers that are uninitialized get `Py_DECREF`'d.. which is probably what is generating the random segfault, and bus errors.

nirosys:
Fixed with #376.

PR #376, opened 2024-10-18
Fix invalid memory access in ionc_write

nirosys:
*Issue #, if available:* #375

*Description of changes:*
The ionc extension had a few uninitialized pointers defined in `ionc_write` that would get `Py_DECREF`'d when the arguments provided where invalid. This was found with the feature added with #372, when an extra parameter was passed to track whether to print trialing commas or not.

This PR initializes the values to NULL and changes the `Py_DECREF` to `Py_XDECREF` because `Py_DECREF` doesn't handle NULL values. Unit tests were also added to ensure we're returning an exception and not blowing up in these cases.

---


nirosys:
Looks like PyPy 3.8 is failing due to ujson and rapidjson extension build issues.. 3.10 pypy on macos looks to be failing because of a linker option in rapidjson.

rmarrowstone (approved):

Issue #305, opened 2023-11-21
DEPRECATION: amazon.ion is being installed using the legacy 'setup.py install' method


lasley:
Performing a `pip install amazon.ion` yields the following deprecation notice

> Installing collected packages: amazon.ion
  DEPRECATION: amazon.ion is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559

jobarr-amzn:
Thanks for bringing this to our attention. I am not personally very familiar with this, from some reading just now in the [Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#) it seems like a `pyproject.toml` would (could?) replace `requirements.txt` and some amount of `setup.cfg` and `setup.py`?

nirosys:
This has been addressed with #379.

PR #379, opened 2024-10-19
Migrate to pyproject.toml

nirosys:
*Issue #, if available:* #305

*Description of changes:*
This PR migrates ion-python away from a setup.py/install.py based project to a [PEP 517](https://peps.python.org/pep-0517/) compliant build system. By updating the build system we are able to adopt better tooling through the formalized nature of PEP 517 projects, and how tooling can integrate with the build tree.

One tool that we can use is a PEP 517 build system called `py-build-cmake` which allows us use CMake for building our C extension. By using CMake, we're able to place ion-c into the build tree and include it as a subdirectory, letting `py-build-cmake` handle the python build flags.

A side effect of this, is that the code layout now splits the C module into `src/` and the python code into `src-python/`.

## Building the Package
Building the extension is now a simple:
```
$ python -m build
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - py-build-cmake~=0.1.8
* Getting build dependencies for sdist...
...
* Building wheel...
...
[100%] Built target _ioncmodule
-- Installing: /var/folders/ld/sktjjrk132z_f7_95g908dxr0000gr/T/tmp4ekl7bjs/staging/amazon/_ioncmodule.so
Successfully built amazon_ion-0.12.0.tar.gz and amazon_ion-0.12.0-cp312-cp312-macosx_13_0_x86_64.whl
(venv) glitch@147dda5e5395 ~/C/ion-python>
```
This will build both the sdist (source distribution) and a binary wheel for the current platform & python version. This can be used to ensure the module compiles, but does not install the package. By default everything is built in isolation, using a virtual environment and pip, everything (including the C module) is built in a temporary directory. 

## C Extension Changes
Prior to this PR when installing the package the C extension would be present under `amazon.ion.ionc`. Due to a limitation in `py-build-cmake` (at the time of implementation) the C extension is now under top-level package, and has been renamed `_ioncmodule`. This is reflected in the simpleion module when testing for the extension's existance. 

## Dependencies
Prior to `pyproject.toml`, it was common practice to put dependencies into a text file called `requirements.txt`. With PEP 517 dependency management has been formalized, along with other metadata that previously could be provided differently depending on which build tools you were actually using.

In `pyproject.toml` we have a section `[project]` which includes a `dependencies` field, containing all of the dependencies needed for the package to work. We also have some optional dependencies that we can install for benchmarking, and/or unit testing. Unit testing and benchmarking are not something that most users would need when installing ion-python, so these dependencies are kept separate in the `[project.optional-dependencies]` section. Each list in this section is a named set of dependencies or extras that can be installed, but are not installed by default.

`pip` supports these optional dependencies through PEP 517, and they can be installed with pip using a command like:

```
$ python -m pip install -e '.[test]'
```

The above command would install the package in editable mode, which only installs the dependencies. Since we reference `test`, it includes the optional dependencies named `test` with the required dependencies. This allows us to keep benchmarking dependencies (like ujson, rapidjson, etc) separate so that our users do not need to install them unless they want to do benchmarking. Likewise, if the user isn't contributing or changing ion-python, there's no need for specific versions of venv and tox, required by ion-python.

## Configuration
All of the 3rd party tooling configuration (eg. tox, pytest, etc) has been moved into the pyproject.toml.

## Additional Changes
In addition to the `pyproject.toml` migration, I added in a new function for the C extension. Ion C has had a version API since Oct 2023, that can return the version and revision. I've added a new method within the `_ioncmodule` package that exposes this functionality:

```python
>>> import amazon._ioncmodule
>>> amazon._ioncmodule.ionc_version()
'v1.1.3 (rev: d61c09a)'
>>>
```
This should make it easier to determine what Ion C fixes/bugs/etc are present in a given ion-python installation.

---


nirosys:
Windows build issue:
```
LINK : fatal error LNK1181: cannot open input file 'm.lib' [C:\Users\runneradmin\AppData\Local\Temp\build-via-sdist-rkldgbwp\amazon_ion-0.13.0\.py-build-cmake_cache\pp310-pypy310_pp73-win_amd64\_ioncmodule.vcxproj]
```

and I missed the performance regression workflow:
```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
```

.github/workflows/main.yml
```diff
@@ -21,9 +21,8 @@ jobs:
 
       - name: Create a virtual environment
         run: git submodule init && git submodule update && python3 -m venv ./venv && . venv/bin/activate
-      - run: pip install --upgrade setuptools
-      - run: pip install -r requirements.txt
-      - run: pip install -e .
+      - run: python -m pip install build virtualenv
```
rmarrowstone, line 24: Why does this action also install `virtualenv` but the `test-windows` does not?

.github/workflows/performance-regression.yml
```diff
@@ -15,6 +15,7 @@ env:
   specs: '{command:read,format:ion_text} {command:write,format:ion_text} {command:read,format:ion_binary} {command:write,format:ion_binary}'
   test_data_id: 'generated-test-data'
   run_cli: 'python amazon/ionbenchmark/ion_benchmark_cli.py'
+  run_cli_new: 'python src-python/amazon/ionbenchmark/ion_benchmark_cli.py'
```
rmarrowstone, line 18: Should we just replace `run_cli` instead of `run_cli_new`?

C_EXTENSION.md
```diff
@@ -76,16 +76,73 @@ b'\xe0\x01\x00\xea\xe9\x81\x83\xd6\x87\xb4\x83abc\xd3\x8a!{'
...
 ```
-                                   ioncmodule.c
-                                        |
-                                        | 
-                                        ↓ 
-Ion C -------> Ion C binaries -----> setup.py ------> C extension -------------------> Ion Python simpleion module
-      compile                                  setup               import ionc module   
+
+The `amazon.ion.simpleion` module then makes use of this extension when it is available to provide more efficient
+Ion parsing. Importing `amazon._ioncmodule` directly can determine if it is available, however simpleion also provides
```
rmarrowstone, line 95: reading _and_ writing, loading _and_ dumping, not just parsing is all i mean.

pyproject.toml
```diff
@@ -0,0 +1,92 @@
...
+    "jsonconversion==0.2.13", # 1.0.1 tightens the requirements for pytest, and 1.0.0 is broken.
+    "pyparsing==3.0.0",
+]
+dynamic = ["version", "description"]
+
+[project.optional-dependencies]
+test = [
+   "pytest==8.3.3"
+]
+benchmarking = [
```
rmarrowstone, line 22: 🙌

src/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/ion-c/decNumber/include>
+   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/ion-c/ionc/include>
+)
+
+install(TARGETS _ioncmodule objlib
+        EXCLUDE_FROM_ALL
+        COMPONENT python_module
+        DESTINATION ${PY_BUILD_CMAKE_MODULE_NAME})
+
+# install(FILES _ioncmodule.pyi
```
rmarrowstone, line 28: delete?

src/ioncmodule.c
```diff
@@ -1531,6 +1531,12 @@ PyObject* ionc_read(PyObject* self, PyObject *args, PyObject *kwds) {
     return exception;
 }
 
+PyObject *ionc_version() {
```
rmarrowstone, line 1534: nice.

rmarrowstone (approved):
Awesome!

.github/workflows/main.yml
nirosys: Ah good catch, ty. I'll remove that, we're not using virtualenv in the workflows.

.github/workflows/performance-regression.yml
nirosys: The new `run_cli_new` is temporary, to allow this PR to run the perf regression workflow. Since it pulls down the current HEAD which has the pre-PR layout, I need to run the benchmark CLI from that revision with `run_cli`.

Once this PR is merged, then I can follow up with a PR to replace `run_cli` with `run_cli_new`, so that the perf regression workflow continues working using the `src-python` layout for everything.

src/CMakeLists.txt
nirosys: Ah! Yes, I'll get rid of that and follow up. There is a way to export python stubs for extensions to allow IDEs to present function and type information for an extension's methods. I left that as a reminder, but will follow up with it after this PR.

.github/workflows/performance-regression.yml
rmarrowstone: Makes sense.

src/CMakeLists.txt
rmarrowstone: That would be neat.

PR #388, opened 2024-11-13
Fix release workflow, externally managed environment error

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
With the release of 0.13.0 a few maintenance issues have popped up with the release workflow. The latest is a change to the MacOS runners which introduces [PEP-668](https://peps.python.org/pep-0668/), disallowing pip to be used directly on a system install. ([Example](https://github.com/amazon-ion/ion-python/actions/runs/11807251033/job/32893620133) failed run)

Most of our workflows use virtual environments, which satisfies PEP-668, however the release workflow does not use a virtual environment when building wheels with cibuildwheel.

This PR updates the release workflow to lean on the cibuildwheel GHA for building our wheels, rather than manually installing and running cibuildwheel. This simplifies the workflow, and allows the cibuildwheel version to be maintained by dependabot. Internally cibuildwheel's GHA uses pipx which satisfies the PEP-668 issue.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,24 +111,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up pip and setuptools
-        run: |
-          pip3 install -U pip
-          pip3 install -U setuptools
-
-      - name: Set up QEMU
-        if: runner.os == 'Linux'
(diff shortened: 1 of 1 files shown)


rmarrowstone (approved):
The best yaml is the yaml one has deleted.

tgregg (approved):

PR #399, opened 2025-01-10
Pulled ion-c submodule to include fix for hanging

### LINKED ISSUE #398: Loads Hangs on Truncated Ion Container
When using the c extension on an Ion Value Stream that ends prematurely within an Ion Container, _but not within a scalar value inside it_, loads hangs indefinitely.

For example:
```
ion.loads(b'\xe0\x01\x00\xea\xb6') # a 6 octet list is declared, stream ends before value is provided
```

premature ends within scalars appears to work fine:
```
ion.loads(b'\xe0\x01\x00\xea!') # a top-level Ion Int is declared, stream ends before value is provided
ion.loads(b'\xe0\x01\x00\xea\xb2!') # value within a 6 octet list ends prematurely itself.
```

It looks like there needs to be a check in `ioncmodule.c` for this. I suspect the root of the problem there is that ion-c returns `tid_EOF` for _both_ EOF and end of each container. So an EOF is expected within a container.

rmarrowstone:
Fixes #398


nirosys (approved):
Nice, looks like two bug fixes with this update.

PR #400, opened 2025-01-14
Update release workflow to use PyPI's trusted publisher support.

nirosys:
*Issue #, if available:* n/a

*Description of changes:*
Prior to this PR the release workflow would build our wheels and upload them to S3, where a code pipeline would upload them to PyPI. Last year, PyPI removed the ability to upload wheels using user/pass for auth, and required either an API key, or configuring a Trusted Publisher.

This PR updates the release workflow to use PyPI's trusted publisher via Github. Rather than uploading the wheels to S3, and uploading them via the code pipeline with a secret key, we've configured our PyPI project to trust Github's OIDC provider. In our workflow we can request a token from Github (via the `pypa/gh-action-pypi-publish` action) which will be signed with Github's key, and verified by PyPI. The token contains claims containing the repo and workflow that is attempting to authenticate. Our PyPI project is configured to only allow the `amazon-ion/ion-python` repo and `release.yml` workflow, to push new wheels. By using this method, we do not have to store any secrets for our PyPI account, and can simplify our release process.

In addition to this change, I've addressed some `actionlint` warnings regarding shell quoting, and added python 3.12 and 3.13 to the versions cibuildwheel will build for.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,7 +71,7 @@ jobs:
           do
             v=""
 
-            if ! (pip install --no-binary :all: amazon.ion==${GITHUB_REF##*/v}) then
+            if ! (pip install --no-binary :all: "amazon.ion==${GITHUB_REF##*/v}") then
               echo "Unable to install the desired version"
               sleep 120s
               continue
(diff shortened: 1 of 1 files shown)


rmarrowstone (approved):
Great!

jobarr-amzn (approved):
Excellent!

tgregg (approved):

PR #431, opened 2026-06-05
Fix release workflow and bump version to 0.14.1

zihaowang678:
The v0.14.0 release failed because the test job installed the package without test dependencies, so pytest was never available. Install with `.[test]` to include pytest.

Also fix the version in `src-python/amazon/__init__.py` which was still at 0.13.0 — this caused the build to produce 0.13.0 artifacts from the v0.14.0 tag.


tgregg (commented):
Looks good to me, but can you also update CHANGES.md with an entry for 0.14.1? It looks like the only change is an ion-c submodule update, so the entry will be short.

tgregg (approved):

PR #445, opened 2026-06-26
Upload sdist to PYPI as well

NovemberZulu:
*Description of changes:*
Currently, source isn't published to PYPI; this is a regression: compare [0.14.5 files](https://pypi.org/project/amazon-ion/0.14.5/#files) and [0.13.0 files](https://pypi.org/project/amazon-ion/0.13.0/#files), the latter has source distribution, the former does not.

The source distribution is built under `source` name, see, e.g. [0.14.5 release artifacts](https://github.com/amazon-ion/ion-python/actions/runs/28200492545#artifacts). It is NOT used in the publishing step, because this step explicitly downloads only `wheel-8` artifacts: https://github.com/amazon-ion/ion-python/blob/a2a13c71b4a9daaec4f7f71fc5266aa28cc89358/.github/workflows/release.yml#L129

This PR changes artifact naming so that sdist is downloaded in the publishing step and  published to PYPI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Upload source to Github
         uses: actions/upload-artifact@v4
         with:
-          name: source
+          name: dist-source
           path: ./dist/amazon_ion-*.tar.gz
 
       - uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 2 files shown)


NovemberZulu:
Updated with better naming and a test for sdist install. The test is modeled after existing wheel test, but is non-blocking

NovemberZulu:
Added changelog entry

popematt (approved):

PR #448, opened 2026-09-09
Report container nesting past the recursion limit as IonException

deirdresama:
*Issue #, if available:*

*Description of changes:*

Deeply nested containers raised `RecursionError`, `SystemError`, or a misreported
ion-c error code depending on the code path taken. All paths now raise
`IonException`, with the original error kept as `__cause__`. Also bumps the
vendored ion-c to v1.1.6.


src-python/amazon/ion/simpleion.py
```diff
@@ -494,16 +510,22 @@ def load_extension(fp, single_value=True, parse_eagerly=True,
...
+            except StopIteration:
+                pass
         return value
     if parse_eagerly:
-        return list(iterator)
+        with _translate_recursion_error():
+            return list(iterator)
+    # The raw iterator is returned so that ionc_write can recognise its type and re-serialize the
+    # stream directly. A caller advancing it past the recursion limit may see a RecursionError
+    # instead of an IonException, depending on which depth limit the runtime reaches first.
```
tgregg: Is the comment accurate? Aren't we ensuring IonException is thrown?

deirdresama: Yes, it's accurate — this is the one remaining path that surfaces a bare
`RecursionError`. I checked all 12 public entry points across 3.9–3.13; the other
11 raise `IonException` everywhere. (This one does too on 3.12+, but only
incidentally: ion-c's `max_container_depth` is reached first there and raises from
the C side.)

The reason it's uncovered is that this branch hands back the raw C iterator, and
the caller drives it with their own `next()` calls, outside anything
`_translate_recursion_error` can wrap. I did try wrapping it in a generator, but
that breaks `ionc_write`'s fast path for re-serializing a stream — it type-checks
for the raw iterator — which showed up as a `test_ion_stream` failure.

Covering it would mean teaching `ionc_write` to accept a wrapped iterator as well,
so it's contained but touches the write path for the sake of one lazy read entry
point. Do you want me to close the gap?

deirdresama: Gap closed.

tgregg (approved):

